PR #1, opened 2017-03-01
CHARMM migration from openmm

ChayaSt:
Migrating https://github.com/pandegroup/openmm/pull/1542


ChayaSt:
What's currently left to do is add the ability to write out the `<patch>` field in parmed. 

(https://github.com/peastman/openmm/blob/238a10688d85c78ea67dc459ea7f5c458b175673/docs-source/usersguide/application.rst#patches)[How patches are implemented in openmm]

The main issue with writing out the patches in parmed is that the documentation in the CHARMM files are not consistent.

jchodera:
@ChayaSt : Could you fix that link and elaborate a bit more? Does ParmEd load the CHARMM `PATC`hes, but doesn't yet write those out when writing OpenMM `ffxml` files?

ChayaSt:
Yes, ParmEd loads the patches, but I don't think it loads which residues it can be applied too which is what's needed for the `<patch>` field. 

In the CHARMM files, the patch usually has some comments about where the patches should be applied (but not always). Sometimes it doesn't name the residue it should be patched too with the CHARMM residue name. See this [example](https://github.com/ChayaSt/Charmm36_FF/blob/master/toppar/top_all22_prot.rtf#L1536)

Sometimes it does name the residue correctly. See [here](https://github.com/ChayaSt/Charmm36_FF/blob/master/toppar/top_all36_carb.rtf#L5268)

jchodera:
@ChayaSt : Let's go ahead and merge this. We can clean things up in subsequent PRs.

Issue #11, opened 2017-04-14
Add setup.py and continuous integration boilerplate


jchodera:
The `setup.py` file will 
* install the conversion scripts (using entry points?)
* install the forcefields into an appropriate package-level directory
* expose the forcefields for OpenMM autodiscovery via an appropriate entry-point
* install input files into `$PREFIX/share/openmm/forcefields/` directory so users can tweak conversion if needed?

jchodera:
cc: pandegroup/openmm#1789

We should define an entry point in the group (currently `openmm.forcefielddir`) that defines a function that returns a string containing a path to the directory that the forcefields are installed in, which would presumably use `pkg_resources.resource_filename` to find where the package is installed.

See http://setuptools.readthedocs.io/en/latest/setuptools.html#dynamic-discovery-of-services-and-plugins

jchodera:
This was done in #26

PR #26, opened 2017-04-17
Add travis continuous integration

jchodera:
Addresses #11


jchodera:
We've at least got AMBER support now!

Once this is merged, I'll start pushing test conda packages.

jchodera:
I'll fix the travis failures in the next PR.

PR #40, opened 2017-10-09
Created script for merging lipids

peastman:


jchodera:
Thanks! I'll clean this up and add an energy comparison test.

jchodera:
Apologies for not having been able to integrate this sooner, @peastman---we had some deadlines to clear through but I'm working to finish this up now.

Issue #34, opened 2017-09-09
Convert non-CHARMM water models


jchodera:
We should convert the water/ion stream files found in the [`non_charmm/`](https://github.com/choderalab/openmm-forcefields/tree/master/charmm/toppar/non_charmm) directory so users can mix-and-match. 

cc: #22

jchodera:
@peastman: This needs to be tackled by extending the YAML syntax used by [`convert_charmm.py`](https://github.com/choderalab/openmm-forcefields/tree/master/charmm) to process multiple sets of input files, and then specify all the water/ion stream combinations in the [`non_charmm/`](https://github.com/choderalab/openmm-forcefields/tree/master/charmm/toppar/non_charmm) directory. Unlike the [Amber conversion scripts](https://github.com/choderalab/openmm-forcefields/tree/master/amber), which can process multiple conversions specified in one YAML file, I think the CHARMM conversion scripts only process a single conversion within a YAML file in their current form.

jchodera:
We will also need to split out the default choices into a separate `ffxml` file and create a meta-package for the default choice and CHARMM parameters.

jchodera:
Merging #42 is currently blocked by waiting for @swails to merge https://github.com/ParmEd/ParmEd/pull/914

jchodera:
https://github.com/ParmEd/ParmEd/pull/914 has been merged, so this can proceed.

jchodera:
These have been converted now that #42 has been merged, though we need to see if we can include [`toppar/stream/misc/toppar_ions_won.str`](https://github.com/choderalab/openmm-forcefields/blob/master/charmm/toppar/stream/misc/toppar_ions_won.str) at some point in the future.

Issue #16, opened 2017-04-14
CHARMM: Ensure correct CHARMM metapackage will be built


jchodera:
We will want to make sure the metapackage includes all files that can be properly converted and combined, while excluding those that contain problematic parameters (until they are fixed by the CHARMM folks).

jchodera:
Since the conversion generates a single XML file, we don't need to have metapackages.

jchodera:
@ChayaSt : I am getting a bunch of warnings like this when I run the conversion script:
```
lski1962:charmm choderaj$ python -i convert_charmm.py 
/Users/choderaj/miniconda3/lib/python3.5/site-packages/parmed/charmm/parameters.py:797: ParameterWarning: Replacing residue 'CHL1'
  resname, ParameterWarning)
/Users/choderaj/miniconda3/lib/python3.5/site-packages/parmed/charmm/parameters.py:797: ParameterWarning: Replacing residue 'CHNS'
  resname, ParameterWarning)
/Users/choderaj/miniconda3/lib/python3.5/site-packages/parmed/charmm/parameters.py:797: ParameterWarning: Replacing residue 'CHM1'
  resname, ParameterWarning)
/Users/choderaj/miniconda3/lib/python3.5/site-packages/parmed/charmm/parameters.py:797: ParameterWarning: Replacing residue 'DMPR'
  resname, ParameterWarning)
/Users/choderaj/miniconda3/lib/python3.5/site-packages/parmed/charmm/parameters.py:797: ParameterWarning: Replacing residue 'MGU1'
  resname, ParameterWarning)
/Users/choderaj/miniconda3/lib/python3.5/site-packages/parmed/charmm/parameters.py:797: ParameterWarning: Replacing residue 'MGU2'
  resname, ParameterWarning)
/Users/choderaj/miniconda3/lib/python3.5/site-packages/parmed/charmm/parameters.py:797: ParameterWarning: Replacing residue 'NIC'
  resname, ParameterWarning)
/Users/choderaj/miniconda3/lib/python3.5/site-packages/parmed/charmm/parameters.py:797: ParameterWarning: Replacing residue 'NICH'
  resname, ParameterWarning)
/Users/choderaj/miniconda3/lib/python3.5/site-packages/parmed/charmm/parameters.py:797: ParameterWarning: Replacing residue 'PPI1'
  resname, ParameterWarning)
[...]

PR #42, opened 2017-10-22
Convert water models separately

jchodera:
Addresses #34 and #16
- [x] Allow script to process multiple forcefield sets from one YAML file
- [x] Add YAML files for non-CHARMM waters
- [x] Split out waters and ions from CHARMM36 conversion
- [x] Create CHARMM36 metapackage including default waters and ions


jchodera:
I'll need to fix [this parmed issue](https://github.com/ParmEd/ParmEd/issues/908) before finishing the splitting of waters.

jchodera:
The major remaining issue (noted [here](https://github.com/choderalab/openmm-forcefields/issues/16#issuecomment-345534553)) seems to be that including the ions from  [`toppar/stream/misc/toppar_ions_won.str`](https://github.com/choderalab/openmm-forcefields/blob/master/charmm/toppar/stream/misc/toppar_ions_won.str) has an ion residue named `Tm3p` that causes a conflict with  [`toppar/top_all36_cgenff.rtf`](https://github.com/choderalab/openmm-forcefields/blob/master/charmm/toppar/top_all36_cgenff.rtf) residue `TM3P` (C5H9O5P 4'-methyl,3'-phosphate tetrahydrofuran (phosphate -2))---both are converted to all-uppercase. For now, I'll exclude the `toppar_ions_won.str` file, but we'll want to find a better long-term solution.

jchodera:
Travis tests are just timing out, though python 3.6 is failing to import `yaml` for some reason I still don't understand.

PR #43, opened 2017-10-28
Add tip3p-fb and tip4p-fb

rafwiewiora:


rafwiewiora:
Failures are due to CHARMM.

jchodera:
Thanks so much for this, @rafwiewiora!

Issue #15, opened 2017-04-14
CHARMM: Automate testing of conversion scripts and resulting converted parameters via travis


jchodera:

jchodera:
@peastman: Unlike the [AMBER conversion script](https://github.com/choderalab/openmm-forcefields/blob/master/amber/convert_amber.py#L506), which validates the conversion by checking energies by component, the [CHARMM conversion script](https://github.com/choderalab/openmm-forcefields/tree/master/charmm) doesn't automatically validate the energy components of some test systems following conversion. 

The functions that @ChayaSt had used were somehow removed during moving the files into `openmm-forcefields`, but you can find the validation code [here](https://github.com/ChayaSt/openmm/commit/520b27a9dbd4a2d3d7a776c35af2cffc0bba93f6) to reintegrate. The test systems @ChayaSt used are [here](https://github.com/ChayaSt/openmm/commit/6f60bf4e4475fd9f2e5b5286ee010716a3df6191#diff-759358248fabfbc0678f532181b8ff84).

jchodera:
Automated CHARMM conversion testing is being added in #48.

PR #48, opened 2017-11-19
Automate testing of CHARMM ffxml files

jchodera:
Addresses #15 : Automate testing of conversion scripts and resulting converted parameters via travis

- [x] Create automated testing script in separate conversion script
- [x] Automate testing on travis
- [ ] Add test systems from CHARMM-GUI: 
  - [ ] vacuum protein
  - [ ] protein+small molecule
  - [ ] protein+membrane
- [ ] Get tests to pass
- [ ] Merge into main CHARMM conversion script


jchodera:
I'm currently running into this issue:
```
lski2414:charmm choderaj$ python test_charmm.py
Testing methanol with ions
/Users/choderaj/miniconda/lib/python3.6/site-packages/ParmEd-2.7.3+283.g4f6eb417.dirty-py3.6-macosx-10.7-x86_64.egg/parmed/openmm/topsystem.py:183: OpenMMWarning: Unsupported Force type CustomNonbondedForce
  warnings.warn('Unsupported Force type %s' % type(force).__name__, OpenMMWarning)
  File "test_charmm.py", line 123, in <module>
    main()
  File "test_charmm.py", line 31, in main
    test_charmm()
  File "test_charmm.py", line 48, in test_charmm
  File "test_charmm.py", line 95, in compare_energies
    omm_energies = openmm.energy_decomposition_system(topology, system_openmm, nrg=units)
  File "/Users/choderaj/miniconda/lib/python3.6/site-packages/ParmEd-2.7.3+283.g4f6eb417.dirty-py3.6-macosx-10.7-x86_64.egg/parmed/utils/decorators.py", line 31, in new_fcn
    return fcn(*args, **kwargs)
  File "/Users/choderaj/miniconda/lib/python3.6/site-packages/ParmEd-2.7.3+283.g4f6eb417.dirty-py3.6-macosx-10.7-x86_64.egg/parmed/openmm/utils.py", line 96, in energy_decomposition_system
    con = mm.Context(system, mm.VerletIntegrator(0.001))
  File "/Users/choderaj/miniconda/lib/python3.6/site-packages/simtk/openmm/openmm.py", line 6265, in __init__
    this = _openmm.new_Context(*args)
Exception: All Forces must have identical exceptions
```
We seem to have two problems:
1. It appears that ParmEd's `energy_decomposition_system` may not support `CustomNonbondedForce` in [`load_topology()`](https://github.com/ParmEd/ParmEd/blob/master/parmed/openmm/topsystem.py#L158-L183), which is used to represent some particular kinds of CHARMM forces. In this case, the serialized XML file suggests that the ParmEd-generated System has a `CustomNonbondedForce` representing LJ interactions:
```XML
<Force cutoff="1" energy="(a/r6)^2-b/r6; r6=r2*r2*r2; r2=r^2; a=acoef(type1, type2); b=bcoef(type1, type2)" forceGroup="11" method="0" switchingDistance="-1" type="CustomNonbondedForce" useLongRangeCorrection="1" useSwitchingFunction="0" version="2">
...
```
2. `ForceField.createSystem()` is somehow creating `System` objects with forces that have different exceptions for `CustomNonbondedForce`:
```XML
                        <Exceptions>
                                <Exception eps="0" p1="0" p2="1" q="0" sig="1"/>
                                <Exception eps="0" p1="0" p2="2" q="0" sig="1"/>
                                <Exception eps="0" p1="1" p2="2" q="0" sig="1"/>
                                <Exception eps="0" p1="0" p2="3" q="0" sig="1"/>
                                <Exception eps="0" p1="1" p2="3" q="0" sig="1"/>
                                <Exception eps="0" p1="2" p2="3" q=".0387" sig="1"/>
                                <Exception eps="0" p1="0" p2="4" q="0" sig="1"/>
                                <Exception eps="0" p1="1" p2="4" q="0" sig="1"/>
                                <Exception eps="0" p1="2" p2="4" q=".0387" sig="1"/>
                                <Exception eps="0" p1="3" p2="4" q="0" sig="1"/>
                                <Exception eps="0" p1="0" p2="5" q="0" sig="1"/>
                                <Exception eps="0" p1="1" p2="5" q="0" sig="1"/>
                                <Exception eps="0" p1="2" p2="5" q=".0387" sig="1"/>
                                <Exception eps="0" p1="3" p2="5" q="0" sig="1"/>
                                <Exception eps="0" p1="4" p2="5" q="0" sig="1"/>
                        </Exceptions>
```
and `NonbondedForce`:
```XML
                        <Exclusions>
                                <Exclusion p1="0" p2="1"/>
                                <Exclusion p1="0" p2="2"/>
                                <Exclusion p1="1" p2="2"/>
                                <Exclusion p1="0" p2="3"/>
                                <Exclusion p1="1" p2="3"/>
                                <Exclusion p1="0" p2="4"/>
                                <Exclusion p1="1" p2="4"/>
                                <Exclusion p1="3" p2="4"/>
                                <Exclusion p1="0" p2="5"/>
                                <Exclusion p1="1" p2="5"/>
                                <Exclusion p1="3" p2="5"/>
                                <Exclusion p1="4" p2="5"/>
                        </Exclusions>
```

jchodera:
It looks like the `NonbondedForce` exceptions that include both charge and LJ exceptions are indeed included in the `CustomNonbondedForce` exceptions, but the LJ-only exceptions are not included. For this methanol example, these LJ exceptions
```XML
 <Exception eps="0" p1="2" p2="3" q=".0387" sig="1"/>
```
records are omitted from `CustomNonbondedForce`.

jchodera:
> Is this being created with a <LennardJonesForce> tag rather than a <CustomNonbondedForce> tag?

The [`charmm36_nowaters.xml`](https://github.com/choderalab/openmm-forcefields/blob/d508b45e4795f6d7c989f09e28811d8332abee39/charmm/ffxml/charmm36_nowaters.xml) file contains the conversion of [`https://github.com/choderalab/openmm-forcefields/blob/d508b45e4795f6d7c989f09e28811d8332abee39/charmm/toppar/par_all36_cgenff.prm`](toppar/par_all36_cgenff.prm), which encodes a `<NonbondedForce>` tag:
```xml
  <NonbondedForce coulomb14scale="1.0" lj14scale="1.0">
    <UseAttributeFromResidue name="charge"/>
    <Atom epsilon="0.0" sigma="1.0" type="HN1"/>
    <Atom epsilon="0.0" sigma="1.0" type="HN2"/>
    <Atom epsilon="0.0" sigma="1.0" type="HN3"/>
...
```
as well as a `<LennardJonesForce>` tag:
```xml
<LennardJonesForce lj14scale="1.0">
    <Atom epsilon="0.192464" sigma="0.04000135244450124" type="HN1"/>
    <Atom epsilon="0.192464" sigma="0.04000135244450124" type="HN2"/>
    <Atom epsilon="0.192464" sigma="0.19599771799087468" type="HN3"/>
```
There is no `<CustomNonbondedForce>` tag.



Here's the PDB file for the [methanol example](https://github.com/choderalab/openmm-forcefields/blob/d508b45e4795f6d7c989f09e28811d8332abee39/charmm/tests/methanol_ions.pdb):
```
CRYST1   12.009   12.338   11.510  90.00  90.00  90.00 P 1           1
ATOM      1  CB  MEOHM   1      -0.748  -0.015   0.024  1.00  0.00      METH C
ATOM      2  OG  MEOHM   1       0.558   0.420  -0.278  1.00  0.00      METH O
ATOM      3  HG1 MEOHM   1       0.716   1.404   0.137  1.00  0.00      METH H
ATOM      4  HB1 MEOHM   1      -1.293  -0.202  -0.901  1.00  0.00      METH H
ATOM      5  HB2 MEOHM   1      -1.263   0.754   0.600  1.00  0.00      METH H
ATOM      6  HB3 MEOHM   1      -0.699  -0.934   0.609  1.00  0.00      METH H
ATOM      7  SOD SOD I   2      -5.844   1.432   3.239  1.00  0.00      ION NA
ATOM      8  CLA CLA I   3      -5.605  -3.153   1.213  1.00  0.00      ION CL
CONECT 1    2    4     5     6
CONECT 2    3
END
END
```

jchodera:
> So we should remove the if and just add the exclusions and bonded force regardless.

@peastman : I've verified this addresses this issue. I'll create a PR for OpenMM. Thanks!

jchodera:
I'm now working to debug the discrepancies in the [methanol test](https://github.com/choderalab/openmm-forcefields/blob/d1b9171d0cd0c0c41e6b54c9dbae0cd861b72648/charmm/test_charmm.py#L46):
```
charmm_energies
[('HarmonicBondForce', 34.560699462890625), ('HarmonicAngleForce', 1.1359474658966064), ('PeriodicTorsionForce', 4.518716812133789), ('HarmonicBondForce', 0.0331103540956974), ('NonbondedForce', -209.5557861328125), ('CustomNonbondedForce', -0.40933549404144287), ('CMMotionRemover', 0.0)]
openmm_energies
[('HarmonicBondForce', 34.59381103515625), ('HarmonicAngleForce', 1.1359474658966064), ('PeriodicTorsionForce', 3.5145576000213623), ('CustomTorsionForce', 0.0), ('NonbondedForce', -207.8372802734375), ('CustomNonbondedForce', 33.115386962890625), ('CMMotionRemover', 0.0), ('CustomBondForce', -0.057463616132736206)]
```
The `charmm_energies` come from ParmEd's `Structure.createSystem()`, while the `openmm_energies` come from `ForceField.createSystem()`. I don't think it will be possible to modify either ParmEd or OpenMM to create `Force`s in the same order, so I'll need to group the forces for comparison.

peastman:
It looks like the big difference is in the CustomNonbondedForces.  Are those supposed to be representing the same thing in both cases?

jchodera:
Yep. The test code is [here](https://github.com/choderalab/openmm-forcefields/blob/d1b9171d0cd0c0c41e6b54c9dbae0cd861b72648/charmm/test_charmm.py#L37-L107). 

I'm currently only testing the one methanol + ions system that @ChayaSt was able to test before patches, since that's the only thing we had on hand at the moment. I'll add more from CHARMM-GUI soon.

peastman:
I'm comparing the CustomNonbondedForces.  The parameters seem to be similar, but not identical.

The CHARMM version defines the energy like this:

```
(a/r6)^2-b/r6; r6=r2*r2*r2; r2=r^2; a=acoef(type1, type2); b=bcoef(type1, type2)
```

And the OpenMM version defines it like this:

```
acoef(type1, type2)/r^12 - bcoef(type1, type2)/r^6;
```

Note that they define `acoef` differently, since the CHARMM version squares it.  Writing out `r6` like that will also cause OpenMM to evaluate it inefficiently.  We should change that to `r^6`.  But anyway.

In the CHARMM version, the first two atoms have types 0 and 1, so the interaction between them has 0.0022063204152320674 for `acoef` (4.867849774669802e-06 for `acoef^2`) and 0.0031579111672633917 for `bcoef`.

In the OpenMM version, the first two atoms have types 2 and 5, so their interaction has 4.263863492755382e-06 for `acoef`, and 0.0027879400705463483 for `bcoef`.

They're similar but not the same.

jchodera:
> Writing out r6 like that will also cause OpenMM to evaluate it inefficiently. We should change that to r^6. But anyway.

This code is generated by ParmEd [here](https://github.com/ParmEd/ParmEd/blob/master/parmed/structure.py#L2726-L2727). File a [ParmEd issue](https://github.com/ParmEd/ParmEd/issues) with a note about why it will escape the lepton expression optimization and we can fix it easily!

> In the CHARMM version, the first two atoms have types 0 and 1, so the interaction between them has 0.0022063204152320674 for acoef (4.867849774669802e-06 for acoef^2) and 0.0031579111672633917 for bcoef.
> In the OpenMM version, the first two atoms have types 2 and 5, so their interaction has 4.263863492755382e-06 for acoef, and 0.0027879400705463483 for bcoef.
> They're similar but not the same.

Could this be a precision issue, or is something more suspicious going on?

It looks suspiciously close to a factor of 2^(1/6) that converts between `rmin` and `sigma`, but not *exactly* that.

peastman:
It's definitely not just a precision issue.  The difference between 0.0027 and 0.0031 is way too big for that.  Do we know which of the two (if either) agrees with the energies CHARMM calculates?

peastman:
I verified that the `acoef` and `bcoef` values in the OpenMM system correspond to the `sigma` and `epsilon` values in the force field definition, so I don't think it's a problem in the implementation of `<LennardJonesForce>`.

jchodera:
ParmEd [tests the energy component agreement](https://github.com/ParmEd/ParmEd/blob/master/test/test_openmm_charmm.py) for the code used to [load CHARMM files through ParmEd](https://github.com/choderalab/openmm-forcefields/pull/48/files#diff-72fa7f5fecfbfec66e0889c9a71ed1c1R87), so I'd be surprised if this was incorrect. That would suggest our writing of parameters to the `ffxml` file [here](https://github.com/jchodera/ParmEd/blob/1580d093585fc32be90ef7d0e8972a818cd3c314/parmed/openmm/parameters.py#L749-L797) is suspect (note that is in [this PR](https://github.com/ParmEd/ParmEd/pull/895) that hasn't yet been merged).

peastman:
I'm trying to trace where the parameter values came from.  Here's the definition for `CT3`, which is the type of the first atom:

```xml
<Atom epsilon="0.32635200000000003" sigma="0.3634866770012585" type="CT3"/>
```

A quick conversion shows what values they came from:

```
>>> 0.32635200000000003/4.184
0.078
>>> 0.3634866770012585*(2**(1/6))
0.40800000000000003
```

Looking in `par_all36_prot.prm`, here are the parameters I find:

```
CT3 0.0 -0.0780 2.040 0.0 -0.01 1.9 ! alkane, 4/98, yin, adm jr.
```

They seem to match, except that rmin is different by a factor of 2, but I assume that's just how they define it.

jchodera:
> They seem to match, except that rmin is different by a factor of 2, but I assume that's just how they define it.

Odd, since that suggests the forcefield writer is not the likely culprit.

Our [test](https://github.com/choderalab/openmm-forcefields/pull/48/files#diff-72fa7f5fecfbfec66e0889c9a71ed1c1R88) currently loads the CHARMM version of the files via ParmEd, but we could also load them through the [OpenMM CHARMM reader](http://docs.openmm.org/latest/userguide/application.html#using-charmm-files). Might be good to do that too.

If needed, we can fire up the [CHARMM docker image](https://github.com/swails/OpenMMEnergyComparisons/tree/master/CHARMM#charmm-docker-image) and compute the energies with CHARMM directly too.

peastman:
I found the problem.  These files use different force fields!  Look right at the top of `methanol_ions.psf` and you'll see:

```
 REMARKS topology toppar/top_all36_cgenff.rtf
```

And the atom types are different: the first atom is `CG331` instead of `CT3`.  So it's comparing CHARMM36 to CGenFF!

jchodera:
CGenFF is part of the CHARMM36 distribution and included in our charmm36.xml package. See charmm/files/master.yaml

Now, if there are two definitions of the same atom type name included in that package from different places, and the parameters don't match, we have a problem!

peastman:
They *aren't* the same atom types.  From top_all36_cgenff.rtf:

```
RESI MEOH          0.00 ! CH4O methanol, adm jr.
GROUP
ATOM CB   CG331   -0.04 !  H11
ATOM OG   OG311   -0.65 !     \
ATOM HG1  HGP1     0.42 ! H12--C1--O1
ATOM HB1  HGA3     0.09 !     /      \
ATOM HB2  HGA3     0.09 !  H13       HO1
ATOM HB3  HGA3     0.09
```

And from toppar_all36_prot_model.str:

```
RESI MEOH        0.00 ! methanol, adm jr.
GROUP  ! order of atoms to match that used in ab initio
ATOM CB   CT3   -0.04 !  H11
ATOM OG   OH1   -0.66 !     \
ATOM HG1  H      0.43 ! H12--C1--O1
ATOM HB1  HA3    0.09 !     /      \
ATOM HB2  HA3    0.09 !  H13       HO1
ATOM HB3 HA3 0.09 !
```

jchodera:
I've dug into this a bit more, and we'll probably need @ChayaSt to help us at this point:
1. Our [`master.yaml`](https://github.com/choderalab/openmm-forcefields/blob/master/charmm/files/charmm36.yaml#L10-L37) definitely includes both `toppar/top_all36_cgenff.rtf` and `toppar/stream/prot/toppar_all36_prot_model.str`
2. Running the conversion script
```
python convert_charmm.py --verbose --in files/charmm36.yaml
```
causes hundreds and hundreds of these warnings to be issues 
```
/Users/choderaj/miniconda/lib/python3.6/site-packages/ParmEd-2.7.3+283.g4f6eb417.dirty-py3.6-macosx-10.7-x86_64.egg/parmed/charmm/parameters.py:790: ParameterWarning: Replacing residue TGUA
  , ParameterWarning)
```
@ChayaSt had previously told me to not worry about these, but I fear these are all cases where the same residue is defined in multiple files and there is no guarantee they are the same parameters. I think the second file is overriding the parameters in the first in the methanol case.

This might require we reconsider our decision to merge all the CHARMM parameters into a single file since it seems like there is no feasible way to do this.

What makes sense here?
1. Split all CHARMM files into their ffxml equivalents
2. Combine CHARMM files into logical groups to minimize collisions (likely requiring we impose some sort of precedence if collisions remain)
3. Combine CHARMM parameters, leave residue definitions split up

peastman:
I've been looking for CHARMM documentation related to this, but I haven't found much.  The one relevant thing I found is this page: http://mackerell.umaryland.edu/warning.shtml

Based on that, perhaps the correct solution is just to make sure CGenFF has the lowest priority, and all other files take precedence over it?  We need to make sure though that it's the only source of duplicate definitions.  If we remove the CGenFF files, we should no longer get any of the warnings you mentioned.

In any case, I think it would be a good idea to consult someone from the CHARMM community and get their feedback on it.

jchodera:
CGenFF is a more modern forcefield, and what if users want their small molecules to be parameterized with that instead? Would we have a separate recipe for just CGenFF?

peastman:
> CGenFF is a more modern forcefield

More modern than what?

jchodera:
> More modern than what?

Here are the references associated with those source parameters:
* `toppar/stream/prot/toppar_all36_prot_model.str` : Journal of Physical Chemistry B, 1998, 102, 3586-3616
* `top_all36_cgenff.rtf` : J. Comput. Chem. 2012, 33, 2451-2468.

peastman:
Those files just define the topologies.  The parameters are in other files.  `par_all36_prot.prm` is dated Feb. 2012.

The warning is pretty explicit: "Never use CGenFF for biological macromolecules."  And it describes how to load the files to ensure the other ones take precedence over CGenFF.

jchodera:
Methanol is not a biological macromolecule.

ChayaSt:
> @ChayaSt had previously told me to not worry about these, but I fear these are all cases where the same residue is defined in multiple files and there is no guarantee they are the same parameters.

Hmmm, I see the parameters aren't alike even though these files are meant to be used together. For example, a small molecule and a protein. 

> More modern than what?
The CGenFF forcefield is Charmm's general forcefield for small molecules which was developed from the protein forcefield (only new parameters were refit). 


We initially planned a separate `ffxml` file for each forcefield but the dependency structure between the files are complex. See [this](https://github.com/ParmEd/ParmEd/issues/561) discussion.

peastman:
> Methanol is not a biological macromolecule.

It's a carbohydrate.  The very next sentence is, "The Protein, Nucleic Acid, Carbohydrate and Lipid force fields are highly optimized for their specific purpose."  Just read what they say.  They're very clear about it.

jchodera:
The contents of `toppar/stream/prot/toppar_all36_prot_model.str` are model compounds meant to mimic protein sidechain analogues used in the development of the older CHARMM22 forcefield:
```
*>> All-hydrogen topology for small model compounds used in the <<
*>> development of the CHARMM22 protein all-hydrogen parameters <<
```
Even if the parameters have been updated, these compounds still utilize the *protein* atom types, whereas the topologies in `toppar/top_all36_cgenff.rtf` use the CGenFF small molecule atom types.

The choice of which one to use should really come down to the application. If you want to benchmark or improve the accuracy of protein parameters by simulating model compounds, much like Michael Shirts did in [this paper](http://aip.scitation.org/doi/abs/10.1063/1.1587119), you want to use the `toppar_all36_prot_model.str` definitions. If you want to actually simulate a box of methanol or methanol binding to a protein, you probably want to use the CGenFF version in `toppar/top_all36_cgenff.rtf`.

So this is a case where the user should really make a decision about which version they'd like to use.

I think the right thing to do here is to split out the protein model compounds into their own file. These are very special-purpose. We can set them to override level 2 so that the user can pull these in on top of the `charmm36.xml` metapackage if they really want to simulate the protein-derived analogues of small molecules instead of small molecules intended in `toppar/top_all36_cgenff.rtf`.

Sound reasonable?

ChayaSt:
> In any case, I think it would be a good idea to consult someone from the CHARMM community and get their feedback on it.

I can reach out to Mackerel again to ask about duplicate residues and when which ones should be used. 

> It's a carbohydrate. The very next sentence is, "The Protein, Nucleic Acid, Carbohydrate and Lipid force fields are highly optimized for their specific purpose." Just read what they say. They're very clear about it.

But this carbohydrate is first defined in the `toppar_all36_prot_model.str` file which is a stream file so I'm not sure it should take precedence in this case.

peastman:
> I can reach out to Mackerel again to ask about duplicate residues and when which ones should be used.

That would be great.

jchodera:
> It's a carbohydrate. The very next sentence is, "The Protein, Nucleic Acid, Carbohydrate and Lipid force fields are highly optimized for their specific purpose." Just read what they say. They're very clear about it.

I'm having trouble locating the file and line you're discussing here. Could you post a pointer?

I don't believe methanol is generally considered to belong to the class of carbohydrates. It's generally just considered a simple alcohol, and in the case of `toppar/stream/prot/toppar_all36_prot_model.str`, the sidechain analogue of serine (SER).

peastman:
http://mackerell.umaryland.edu/warning.shtml

jchodera:
That text is specifically warning the user not to try to parameterize a protein or nucleic acid with CGenFF atom types.

jchodera:
From Alex MacKerell in a response to @ChayaSt's question today:
>     CGenFF was created after the majority of small model compounds were originally created, with those model compounds acting as the foundation of CGenFF.  Thus, there will be repeats.  However, the parameters should generally be identical.  It appears that there was divergence in 2006 in the alcohols associated with changes in the carbohydrates and these are what ended up in CGenFF.  So, my suggestion is that you use what's in CGenFF as that is, well, CGenFF.   It's less than ideal that there is a small difference in the alcohol charges between CGenFF/Carbohydrates and Protein/nucleic acids, but there is not much we can do about it now without undertaking a significant amount of work doing some reoptimization and validation of major aspects of the additive FF.   
> Note that the original MeOH parameters used in the proteins and nucleic acids were optimized in ca. 1990, so you are seeing the outcome of years of work on a very wide range of molecular systems.   Mistakes happen....

To me, this suggests we should go with the plan I proposed earlier:
> I think the right thing to do here is to split out the protein model compounds into their own file. These are very special-purpose. We can set them to override level 2 so that the user can pull these in on top of the `charmm36.xml metapackage` if they really want to simulate the protein-derived analogues of small molecules instead of small molecules intended in `toppar/top_all36_cgenff.rtf`.

OK with you @peastman?

peastman:
Sounds good.

jchodera:
Great. I think I can sweep up these next tasks for CHARMM and AMBER tonight:
* AMBER: Use @peastman's updated `{POPC,POPE}.pdb` in test of merged lipid ffxml
* CHARMM: Split out model compounds in `toppar/stream/prot/toppar_all36_prot_model.str` and keep adding test systems

@peastman : Any chance you'd be able to improve the torsion assignment speed performance in `ForceField`?

peastman:
I've been looking into it.  I'm just concerned about not making things worse in other cases.

Currently it has a double loop over all torsions in the topology and all definitions in the force field.  Since the latter is very large, it's slow.  The obvious solution would be to use some sort of lookup table to only loop over a subset of definitions.  I could pick one position, and for every atom type, create a list of all definitions that can have that type at the specified position.  But each position can be a single atom type, a class that corresponds to multiple types, or a wildcard that matches anything.  Plus I have to consider both forward and backward versions of each definition.  So each one of those lists could be quite long, and building them could be slow and use a lot of memory.  But I'll see what I can do.

jchodera:
For an `i,j,k,l` torsion, are the central atoms (`j` and `k`) ever wildcards? 

If not, I imagine this would be fast:
* look up the central atoms `(j,k)` in a dict containing both `(j,k)` and `(k,j)` entries to see if you have a special case, and if not, 
* iterate over wildcards to see which one applies

peastman:
> For an i,j,k,l torsion, are the central atoms (j and k) ever wildcards?

We don't restrict them from being, so they easily could be.  And they'll often be classes that correspond to many atom types.  That's the case for our existing force fields.

jchodera:
@peastman : I've split out the model protein sidechain analogues.

There seems to be an issue with the residues containing the wrong types still.

From `top_all36_cgenff.rtf`:
```
RESI MEOH          0.00 ! CH4O methanol, adm jr.
GROUP
ATOM CB   CG331   -0.04 !  H11
ATOM OG   OG311   -0.65 !     \
ATOM HG1  HGP1     0.42 ! H12--C1--O1
ATOM HB1  HGA3     0.09 !     /      \
ATOM HB2  HGA3     0.09 !  H13       HO1
ATOM HB3  HGA3     0.09
BOND CB  OG  OG HG1
BOND CB HB1  CB HB2  CB  HB3
DONO HG1 OG
ACCE OG
! only for analysis
IC HG1  OG   CB   HB1   0.0000  0.0000  180.0000  0.0000  0.0000
IC OG   HB1  *CB  HB2   0.0000  0.0000  120.0000  0.0000  0.0000
IC OG   HB1  *CB  HB3   0.0000  0.0000 -120.0000  0.0000  0.0000
```
but our `charmm36_nowaters.xml` file contains:
```XML
    <Residue name="MEOH">
      <Atom charge="-0.04" name="CB" type="CC331"/>
      <Atom charge="-0.65" name="OG" type="OC311M"/>
      <Atom charge="0.42" name="HG1" type="HCP1"/>
      <Atom charge="0.09" name="HB1" type="HCA3"/>
      <Atom charge="0.09" name="HB2" type="HCA3"/>
      <Atom charge="0.09" name="HB3" type="HCA3"/>
      <Bond atomName1="CB" atomName2="OG"/>
      <Bond atomName1="OG" atomName2="HG1"/>
      <Bond atomName1="CB" atomName2="HB1"/>
      <Bond atomName1="CB" atomName2="HB2"/>
      <Bond atomName1="CB" atomName2="HB3"/>
    </Residue>
```
Note that this is different from what appears in the `toppar/stream/prot/toppar_all36_prot_model.str` file:
```
RESI MEOH        0.00 ! methanol, adm jr.
GROUP  ! order of atoms to match that used in ab initio
ATOM CB   CT3   -0.04 !  H11
ATOM OG   OH1   -0.66 !     \
ATOM HG1  H      0.43 ! H12--C1--O1
ATOM HB1  HA3    0.09 !     /      \
ATOM HB2  HA3    0.09 !  H13       HO1
ATOM HB3  HA3    0.09 !
BOND CB  OG  OG HG1  
BOND CB HB1  CB HB2  CB  HB3
```
@peastman: If you have a chance to poke around with this, please do! I'm mostly out of commission until Thursday.

jchodera:
I think there are a fair number of duplicates left to cull. See the attached list of current warnings:
[warnings.txt](https://github.com/choderalab/openmm-forcefields/files/1494623/warnings.txt)

@ChayaSt : Do we need to pull these out? If so, how can we find which files have these collisions?

peastman:
I did a search for "RESI MEOH" in the CHARMM data files.  It turns out there are five different files with definitions of it: top_all36_cgenff.rtf, toppar_all36_prot_model.str, toppar_all36_carb_model.str, toppar_drude_model_2013c.str, and top_all30_cheq_prot.inp.

Two of those are from different force fields, so they'll be converted separately from the others, not merged in.  And we're leaving out the model compounds.  That still leaves two definitions of it, one in CGenFF and one in carbohydrates.

jchodera:
How about I just split out all the `_model_` compounds (both CGenFF and carbohydrates) into separate files?

peastman:
That makes sense.  And presumably do the same thing with drude as well.

jchodera:
We had elected to punt on the conversion of Drude for this release when we started this process way back when because ParmEd doesn't support it either.

peastman:
But that's one of the ones users are specifically waiting for: https://github.com/pandegroup/openmm/issues/1611#issuecomment-319138656

jchodera:
We don't have the infrastructure to convert the polarizable forcefield right now, and we can't keep expanding the scope or we'll never finish anything.

I honestly didn't even know the simtk forums were still in existence. I must not have clicked on that link and presumed that people were just really excited for the CHARMM36 modern fixed-charge forcefield.

jchodera:
If you want to take a stab at adding the infrastructure to ParmEd, @peastman, you're welcome to, but that's probably a separate project.

jchodera:
It looks like even OpenMM doesn't have the capability of loading in CHARMM files with Drude oscillators via `app.CharmmParameterSet`. Could that be an alternative way to get some CHARMM Drude functionality into OpenMM if they go through the [CHARMM-GUI Drude Prepper](http://www.charmm-gui.org/?doc=input/drude)? Implementing the ability to read the parameter set into [`charmmparameterset.py`](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/charmmparameterset.py) would also seem like a requisite first step to getting that functionality into ParmEd.

jchodera:
We're getting closer!
```
lski2414:charmm choderaj$ python test_charmm.py 
Testing methanol with ions
/Users/choderaj/miniconda/lib/python3.6/site-packages/ParmEd-2.7.3+289.g4ef07f7e.dirty-py3.6-macosx-10.7-x86_64.egg/parmed/openmm/topsystem.py:183: OpenMMWarning: Unsupported Force type CustomNonbondedForce
  warnings.warn('Unsupported Force type %s' % type(force).__name__, OpenMMWarning)
/Users/choderaj/miniconda/lib/python3.6/site-packages/ParmEd-2.7.3+289.g4ef07f7e.dirty-py3.6-macosx-10.7-x86_64.egg/parmed/openmm/topsystem.py:183: OpenMMWarning: Unsupported Force type CustomBondForce
  warnings.warn('Unsupported Force type %s' % type(force).__name__, OpenMMWarning)
CHARMM total energy: -169.716648
[('HarmonicBondForce', 34.560699462890625), ('HarmonicAngleForce', 1.1359474658966064), ('PeriodicTorsionForce', 4.518716812133789), ('HarmonicBondForce', 0.0331103540956974), ('NonbondedForce', -209.5557861328125), ('CustomNonbondedForce', -0.40933549404144287), ('CMMotionRemover', 0.0)]

OPENMM total energy: -136.198530
[('HarmonicBondForce', 34.59381103515625), ('HarmonicAngleForce', 1.1359474658966064), ('PeriodicTorsionForce', 4.518716812133789), ('CustomTorsionForce', 0.0), ('NonbondedForce', -209.49832153320312), ('CustomNonbondedForce', 33.10877990722656), ('CMMotionRemover', 0.0), ('CustomBondForce', -0.057463616132736206)]
TOTAL ERROR: 33.518118
```
The difference is presumably this 33.1 kcal/mol `CustomNonbondedForce` in OpenMM that is -0.409 kcal/mol in the ParmEd version that is reading the CHARMM files directly.

I've attached the serialized `System` XML files for inspection. Looking through them now.
[system_xml.zip](https://github.com/choderalab/openmm-forcefields/files/1503560/system_xml.zip)

jchodera:
That's odd...the number of bonds is different.

From the OpenMM version:
```XML
		<Force forceGroup="0" type="HarmonicBondForce" usesPeriodic="0" version="2">
			<Bonds>
				<Bond d=".142" k="358150.3999999999" p1="0" p2="1"/>
				<Bond d=".1111" k="269449.6" p1="0" p2="3"/>
				<Bond d=".1111" k="269449.6" p1="0" p2="4"/>
				<Bond d=".1111" k="269449.6" p1="0" p2="5"/>
				<Bond d=".096" k="456055.99999999994" p1="1" p2="2"/>
				<Bond d=".18020000000000003" k="4518.72" p1="3" p2="4"/>
				<Bond d=".18020000000000003" k="4518.72" p1="3" p2="5"/>
				<Bond d=".18020000000000003" k="4518.72" p1="4" p2="5"/>
			</Bonds>
		</Force>
```
and from the ParmEd version:
```XML
		<Force forceGroup="0" type="HarmonicBondForce" usesPeriodic="0" version="2">
			<Bonds>
				<Bond d=".142" k="358150.4" p1="0" p2="1"/>
				<Bond d=".1111" k="269449.60000000003" p1="0" p2="3"/>
				<Bond d=".1111" k="269449.60000000003" p1="0" p2="4"/>
				<Bond d=".1111" k="269449.60000000003" p1="0" p2="5"/>
				<Bond d=".096" k="456056.00000000006" p1="1" p2="2"/>
			</Bonds>
		</Force>
```

jchodera:
OpenMM seems to be adding harmonic bonds between the methyl hydrogens:
```XML
				<Bond d=".18020000000000003" k="4518.72" p1="3" p2="4"/>
				<Bond d=".18020000000000003" k="4518.72" p1="3" p2="5"/>
				<Bond d=".18020000000000003" k="4518.72" p1="4" p2="5"/>
```
The PDB file seems to be correct in specifying `CONECT` records, however:
```
CRYST1   12.009   12.338   11.510  90.00  90.00  90.00 P 1           1
ATOM      1  CB  MEOHM   1      -0.748  -0.015   0.024  1.00  0.00      METH C
ATOM      2  OG  MEOHM   1       0.558   0.420  -0.278  1.00  0.00      METH O
ATOM      3  HG1 MEOHM   1       0.716   1.404   0.137  1.00  0.00      METH H
ATOM      4  HB1 MEOHM   1      -1.293  -0.202  -0.901  1.00  0.00      METH H
ATOM      5  HB2 MEOHM   1      -1.263   0.754   0.600  1.00  0.00      METH H
ATOM      6  HB3 MEOHM   1      -0.699  -0.934   0.609  1.00  0.00      METH H
ATOM      7  SOD SOD I   2      -5.844   1.432   3.239  1.00  0.00      ION NA
ATOM      8  CLA CLA I   3      -5.605  -3.153   1.213  1.00  0.00      ION CL
CONECT 1    2    4     5     6
CONECT 2    3
END
END
```
I wonder where the extra bonds are coming from?

jchodera:
These bonds are not present in `pdbfile.topology`:
```
OpenMM pdbfile.topology bonds:
Bond(<Atom 0 (CB) of chain 0 residue 0 (MEOH)>, <Atom 1 (OG) of chain 0 residue 0 (MEOH)>)
Bond(<Atom 0 (CB) of chain 0 residue 0 (MEOH)>, <Atom 3 (HB1) of chain 0 residue 0 (MEOH)>)
Bond(<Atom 0 (CB) of chain 0 residue 0 (MEOH)>, <Atom 4 (HB2) of chain 0 residue 0 (MEOH)>)
Bond(<Atom 0 (CB) of chain 0 residue 0 (MEOH)>, <Atom 5 (HB3) of chain 0 residue 0 (MEOH)>)
Bond(<Atom 1 (OG) of chain 0 residue 0 (MEOH)>, <Atom 2 (HG1) of chain 0 residue 0 (MEOH)>)
```
Somehow `app.ForceField` is adding them when instantiated with `constraints=None`.

peastman:
Those are probably coming from the `<AmoebaUreyBradleyForce>`.  Originally that was actually a force class, but now it just adds bonds to the standard HarmonicBondForce.

jchodera:
Excellent point! It looks like the OpenMM CHARMM reader is in agreement with `ForceField` now:
```
OpenMM CHARMM reader total energy: -136.198530
ParmEd CHARMM reader total energy: -169.716648
OPENMM total energy: -136.198530
```
For some reason, the ParmEd CHARMM reader must be neglecting the Urey-Bradley force.

jchodera:
The Urey-Bradleys are a red herring, since ParmEd just puts these into a different force group:
```XML
		<Force forceGroup="3" type="HarmonicBondForce" usesPeriodic="0" version="2">
			<Bonds>
				<Bond d=".18020000000000003" k="4518.72" p1="3" p2="5"/>
				<Bond d=".18020000000000003" k="4518.72" p1="3" p2="4"/>
				<Bond d=".18020000000000003" k="4518.72" p1="4" p2="5"/>
			</Bonds>
		</Force>
```
I think the `CustomNonbondedForce` discrepancy is the major one to track down.

ParmEd and OpenMM encode these in a rather different way. The ordering of particles/types is different, but the exclusion lists appear to have the same contents.

ParmEd has, as observed earlier, an encoding that doesn't require the particle pair parameters to have as large of a dynamic range:
```
		<Force cutoff="1" energy="(a/r6)^2-b/r6; r6=r2*r2*r2; r2=r^2; a=acoef(type1, type2); b=bcoef(type1, type2)" forceGroup="11" method="0" switchingDistance="-1" type="CustomNonbondedForce" useLongRangeCorrection="1" useSwitchingFunction="0" version="2">
			<PerParticleParameters>
				<Parameter name="type"/>
			</PerParticleParameters>
			<GlobalParameters/>
			<EnergyParameterDerivatives/>
			<Particles>
				<Particle param1="0"/>
				<Particle param1="1"/>
				<Particle param1="2"/>
				<Particle param1="3"/>
				<Particle param1="3"/>
				<Particle param1="3"/>
				<Particle param1="4"/>
				<Particle param1="5"/>
			</Particles>
			<Exclusions>
				<Exclusion p1="1" p2="0"/>
				<Exclusion p1="2" p2="0"/>
				<Exclusion p1="3" p2="1"/>
				<Exclusion p1="4" p2="1"/>
				<Exclusion p1="5" p2="1"/>
				<Exclusion p1="2" p2="1"/>
				<Exclusion p1="3" p2="0"/>
				<Exclusion p1="4" p2="3"/>
				<Exclusion p1="5" p2="3"/>
				<Exclusion p1="4" p2="0"/>
				<Exclusion p1="5" p2="4"/>
				<Exclusion p1="5" p2="0"/>
				<Exclusion p1="2" p2="3"/>
				<Exclusion p1="2" p2="4"/>
				<Exclusion p1="2" p2="5"/>
			</Exclusions>
			<Functions>
				<Function name="acoef" type="Discrete2DFunction" version="1" xsize="6" ysize="6">
					<Values>
						<Value v=".002713602805589849"/>
						<Value v=".0022063204152320674"/>
						<Value v="6.931457959624168e-05"/>
						<Value v=".00064575903057199"/>
						<Value v=".000864239406207876"/>
						<Value v=".004372650682138487"/>
						<Value v=".0022063204152320674"/>
						<Value v=".001734633592932983"/>
						<Value v="3.8889360171332686e-05"/>
						<Value v=".0004776393697870706"/>
						<Value v=".0006464705277153553"/>
						<Value v=".0036371543219201274"/>
						<Value v="6.931457959624168e-05"/>
						<Value v="3.8889360171332686e-05"/>
						<Value v="3.5946184559525344e-09"/>
						<Value v="5.467534236452303e-06"/>
						<Value v="8.429143614654752e-06"/>
						<Value v=".0001420395083116961"/>
						<Value v=".00064575903057199"/>
						<Value v=".0004776393697870706"/>
						<Value v="5.467534236452303e-06"/>
						<Value v=".00011741138821223224"/>
						<Value v=".00016231123746112688"/>
						<Value v=".0011089569180939415"/>
						<Value v=".000864239406207876"/>
						<Value v=".0006464705277153553"/>
						<Value v="8.429143614654752e-06"/>
						<Value v=".00016231123746112688"/>
						<Value v=".00022349231131320925"/>
						<Value v=".001597951518444677"/>
						<Value v=".004372650682138487"/>
						<Value v=".0036371543219201274"/>
						<Value v=".0001420395083116961"/>
						<Value v=".0011089569180939415"/>
						<Value v=".001597951518444677"/>
						<Value v=".006937078110176043"/>
					</Values>
				</Function>
				<Function name="bcoef" type="Discrete2DFunction" version="1" xsize="6" ysize="6">
					<Values>
						<Value v=".0031004120385176633"/>
						<Value v=".0031579111672633917"/>
						<Value v="6.940064962760784e-05"/>
						<Value v=".0005495064354353205"/>
						<Value v=".0008695143617642285"/>
						<Value v=".005883247017089453"/>
						<Value v=".0031579111672633917"/>
						<Value v=".003110264108099786"/>
						<Value v="4.877844267426442e-05"/>
						<Value v=".0005091674142953035"/>
						<Value v=".0008147973650467806"/>
						<Value v=".006130448068355026"/>
						<Value v="6.940064962760784e-05"/>
						<Value v="4.877844267426442e-05"/>
						<Value v="3.1539699356839008e-09"/>
						<Value v="4.077177445838526e-06"/>
						<Value v="7.431764780585461e-06"/>
						<Value v=".0001674739511329001"/>
						<Value v=".0005495064354353205"/>
						<Value v=".0005091674142953035"/>
						<Value v="4.077177445838526e-06"/>
						<Value v="7.441177726060386e-05"/>
						<Value v=".00012162432514630177"/>
						<Value v=".001111261154564739"/>
						<Value v=".0008695143617642285"/>
						<Value v=".0008147973650467806"/>
						<Value v="7.431764780585461e-06"/>
						<Value v=".00012162432514630177"/>
						<Value v=".00019800440033366964"/>
						<Value v=".0018932401192912343"/>
						<Value v=".005883247017089453"/>
						<Value v=".006130448068355026"/>
						<Value v=".0001674739511329001"/>
						<Value v=".001111261154564739"/>
						<Value v=".0018932401192912343"/>
						<Value v=".010991274335347042"/>
					</Values>
				</Function>
			</Functions>
			<InteractionGroups/>
		</Force>
```
while OpenMM has:
```XML
<Force cutoff="1" energy="acoef(type1, type2)/r^12 - bcoef(type1, type2)/r^6;" forceGroup="0" method="0" switchingDistance="-1" type="CustomNonbondedForce" useLongRangeCorrection="1" useSwitchingFunction="0" version="2">
			<PerParticleParameters>
				<Parameter name="type"/>
			</PerParticleParameters>
			<GlobalParameters/>
			<EnergyParameterDerivatives/>
			<Particles>
				<Particle param1="2"/>
				<Particle param1="1"/>
				<Particle param1="5"/>
				<Particle param1="4"/>
				<Particle param1="4"/>
				<Particle param1="4"/>
				<Particle param1="3"/>
				<Particle param1="0"/>
			</Particles>
			<Exclusions>
				<Exclusion p1="0" p2="1"/>
				<Exclusion p1="0" p2="2"/>
				<Exclusion p1="1" p2="2"/>
				<Exclusion p1="0" p2="3"/>
				<Exclusion p1="1" p2="3"/>
				<Exclusion p1="2" p2="3"/>
				<Exclusion p1="0" p2="4"/>
				<Exclusion p1="1" p2="4"/>
				<Exclusion p1="2" p2="4"/>
				<Exclusion p1="3" p2="4"/>
				<Exclusion p1="0" p2="5"/>
				<Exclusion p1="1" p2="5"/>
				<Exclusion p1="2" p2="5"/>
				<Exclusion p1="3" p2="5"/>
				<Exclusion p1="4" p2="5"/>
			</Exclusions>
			<Functions>
				<Function name="acoef" type="Discrete2DFunction" version="1" xsize="6" ysize="6">
					<Values>
						<Value v="4.8123052706683656e-05"/>
						<Value v="1.3228891561462269e-05"/>
						<Value v="1.9120073988006182e-05"/>
						<Value v=".010458927330507366"/>
						<Value v="1.229785446188412e-06"/>
						<Value v="2.0175221921428418e-08"/>
						<Value v="1.3228891561462269e-05"/>
						<Value v="3.0089537017315934e-06"/>
						<Value v="4.867849774669811e-06"/>
						<Value v="4.1792414320457007e-07"/>
						<Value v="2.2813936757059036e-07"/>
						<Value v="1.5123823345356399e-09"/>
						<Value v="1.9120073988006182e-05"/>
						<Value v="4.867849774669811e-06"/>
						<Value v="7.3636401865051094e-06"/>
						<Value v="7.469097512425411e-07"/>
						<Value v="4.1700472556527737e-07"/>
						<Value v="4.8045109446037345e-09"/>
						<Value v=".010458927330507366"/>
						<Value v="4.1792414320457007e-07"/>
						<Value v="7.469097512425411e-07"/>
						<Value v="4.994881321612037e-08"/>
						<Value v="2.634493780616233e-08"/>
						<Value v="7.105046207647499e-11"/>
						<Value v="1.229785446188412e-06"/>
						<Value v="2.2813936757059036e-07"/>
						<Value v="4.1700472556527737e-07"/>
						<Value v="2.634493780616233e-08"/>
						<Value v="1.3785434081923518e-08"/>
						<Value v="2.989393062677808e-11"/>
						<Value v="2.0175221921428418e-08"/>
						<Value v="1.5123823345356399e-09"/>
						<Value v="4.8045109446037345e-09"/>
						<Value v="7.105046207647499e-11"/>
						<Value v="2.989393062677808e-11"/>
						<Value v="1.2921281843874588e-17"/>
					</Values>
				</Function>
				<Function name="bcoef" type="Discrete2DFunction" version="1" xsize="6" ysize="6">
					<Values>
						<Value v=".010991274335347046"/>
						<Value v=".0061304480683550275"/>
						<Value v=".005883247017089454"/>
						<Value v=".12116736763463902"/>
						<Value v=".0011112611545647387"/>
						<Value v=".00016747395113290024"/>
						<Value v=".0061304480683550275"/>
						<Value v=".0031102641080997875"/>
						<Value v=".0031579111672633948"/>
						<Value v=".0008147973650467807"/>
						<Value v=".0005091674142953039"/>
						<Value v="4.8778442674264466e-05"/>
						<Value v=".005883247017089454"/>
						<Value v=".0031579111672633948"/>
						<Value v=".003100412038517665"/>
						<Value v=".000869514361764228"/>
						<Value v=".0005495064354353212"/>
						<Value v="6.940064962760791e-05"/>
						<Value v=".12116736763463902"/>
						<Value v=".0008147973650467807"/>
						<Value v=".000869514361764228"/>
						<Value v=".0001980044003336695"/>
						<Value v=".0001216243251463018"/>
						<Value v="7.431764780585463e-06"/>
						<Value v=".0011112611545647387"/>
						<Value v=".0005091674142953039"/>
						<Value v=".0005495064354353212"/>
						<Value v=".0001216243251463018"/>
						<Value v="7.44117772606039e-05"/>
						<Value v="4.077177445838527e-06"/>
						<Value v=".00016747395113290024"/>
						<Value v="4.8778442674264466e-05"/>
						<Value v="6.940064962760791e-05"/>
						<Value v="7.431764780585463e-06"/>
						<Value v="4.077177445838527e-06"/>
						<Value v="3.1539699356839016e-09"/>
					</Values>
				</Function>
			</Functions>
			<InteractionGroups/>
		</Force>
```

peastman:
Many of the parameters match up, but not all of them.  For example, in the ParmEd version the last two atoms have types 4 and 5 respectively.  Looking in the `bcoef` table for those types we find the value 0.0018932401192912343.  In the OpenMM version the last two atoms have the types 3 and 0.  Looking in the `bcoef` table we find 0.12116736763463902.  That value doesn't appear anywhere in the ParmEd version, and is way bigger than any other.

peastman:
I think I found the problem.  It's converting the NBFIX terms incorrectly.  For `NONBONDED` terms, CHARMM files specify sigma/2, but for `NBFIX` terms they specify sigma (not divided by 2).  The `<NBFixPair>` tags have sigma twice as large as it should be.

jchodera:
Thanks! Looks like we also have to fix this in OpenMM's `charmmpsffile` in [this line](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/charmmpsffile.py#L1044). Oof---I wonder how we missed that in validating OpenMM 7 against CHARMM energies.

jchodera:
Hm, correcting the NBFix writing in ParmEd's ffxml writer didn't seem to have an effect.

Looking back at the original [`charmm36_nowaters.xml`](https://github.com/choderalab/openmm-forcefields/blob/charmm/charmm/ffxml/charmm36_nowaters.xml) file, there's only a handful of NBFix terms, and they don't involve atom types in our [methanol example](https://github.com/choderalab/openmm-forcefields/blob/charmm/charmm/tests/methanol_ions.psf):
```XML
    <NBFixPair epsilon="0.1343064" sigma="0.5969021411540274" type1="CRL1" type2="HAL2"/>
    <NBFixPair epsilon="0.15020560000000002" sigma="0.4775217129232219" type1="HAL2" type2="HGA1"/>
    <NBFixPair epsilon="0.173636" sigma="0.598683938590308" type1="CRL2" type2="HAL2"/>
    <NBFixPair epsilon="0.1326328" sigma="0.4775217129232219" type1="HAL2" type2="HGA2"/>
    <NBFixPair epsilon="0.1225912" sigma="0.5969021411540274" type1="CRL1" type2="HAL1"/>
    <NBFixPair epsilon="0.1376536" sigma="0.4775217129232219" type1="HAL1" type2="HGA1"/>
    <NBFixPair epsilon="0.1585736" sigma="0.598683938590308" type1="CRL2" type2="HAL1"/>
    <NBFixPair epsilon="0.1209176" sigma="0.4775217129232219" type1="HAL1" type2="HGA2"/>
    <NBFixPair epsilon="0.11506000000000001" sigma="0.7145007719485521" type1="CRL1" type2="CTL1"/>
    <NBFixPair epsilon="0.12886720000000002" sigma="0.5951203437177467" type1="CTL1" type2="HGA1"/>
    <NBFixPair epsilon="0.148532" sigma="0.7162825693848327" type1="CRL2" type2="CTL1"/>
    <NBFixPair epsilon="0.1133864" sigma="0.5951203437177467" type1="CTL1" type2="HGA2"/>
    <NBFixPair epsilon="0.8631592000000001" sigma="0.5026985106978679" type1="MG" type2="SO"/>
    <NBFixPair epsilon="0.8414024" sigma="0.5043912182623345" type1="MG" type2="SS1"/>
```

jchodera:
Whoops! It's the [water file](https://github.com/choderalab/openmm-forcefields/blob/charmm/charmm/ffxml/waters_ions_default.xml#L126-L136) that contains the relevant NBFix terms between sodium and chloride:
```XML
    <NBFixPair epsilon="0.35093300000000005" sigma="0.6647886234763212" type1="CLA" type2="SOD"/>
```
We have one of each of those in our methanol test system.

jchodera:
I had forgotten to re-run the water conversions. When I do, this, everything matches up for methanol!
```
OpenMM CHARMM reader total energy: -169.716644
ParmEd CHARMM reader total energy: -169.716648
OPENMM ffxml total energy: -169.716644
TOTAL ERROR: 0.000003
```
It looks like we do *not* have to change the OpenMM CHARMM reader NBFix. I had a typo in my test script that was showing me the wrong energy for that route.

peastman:
😀

jchodera:
@peastman : I'm going to go ahead and merge this once I've cleaned up travis a bit.

I think we can proceed with rolling files into OpenMM for testing at this point as I implement more automated integration tests into this repo in parallel with manual testing.

Major question: What should I be automating integrated system testing of `ForceField` against?
1. OpenMM's CHARMM reader
2. ParmEd's CHARMM reader
3. [Free CHARMM](https://www.charmm.org/charmm/showcase/news/free-charmm/) in a docker container, as [we did for OpenMM 7](https://github.com/swails/OpenMMEnergyComparisons)

Currently, we're doing (1) and (2). Is this sufficient?

peastman:
If we can do 3 as well that would be great.  We certainly need to test against CHARMM itself one way or another, whether the tests are automated or run manually.

jchodera:
> If we can do 3 as well that would be great. We certainly need to test against CHARMM itself one way or another, whether the tests are automated or run manually.

OK, that should be possible.

PR #49, opened 2017-11-19
Add AMBER ff15ipq using unreleased parm15ipq_10.3.dat update

jchodera:
This PR hacks in AMBER ff15ipq support using an unreleased `parm15ipq_10.3.dat` update from:

https://github.com/choderalab/openmm-forcefields/issues/44#issuecomment-343880821

cc: https://github.com/choderalab/openmm-forcefields/issues/20#issuecomment-345553915


dscerutti:
Thanks, John.  The file I gave you should give identical energies to the release version if you convert both into OpenMM format and then run your tests.  The new parm15ipq_10.3.dat, and the OpenMM conversion of the release ff15ipq containing the earlier parm file, should give slightly different energies when compared to topologies generated by Amber tleap, but Amber tleap utilizing the new parm file should fall back in line and be consistent.

Please let me know once you make this test and confirm that it was successful.  I will note a success as part of the validation documentation when I do the patch.

jchodera:
@dscerutti : I think the only question for us is whether the file you provided us [`parm15ipq_10.3.zip`](https://github.com/choderalab/openmm-forcefields/files/1466469/parm15ipq_10.3.zip) is identical to the final release version; @peastman assumed it would be. If not, he'll have to fix that.

For the energy validation, we compare to LEaP generated `prmtop/inpcrd` files (generated with the same updated `parm15ipq_10.3.zip` you sent us) loaded into OpenMM.

I'm not quite sure I follow your note above---apologies for my confusion!

jchodera:
This PR is working as intended, so I'm going to merge it now.

dscerutti:
The file I included in the .zip folder is intended to be the final release version.

peastman:
Ok, great.  If for any reason it changes, let us know right away.

jchodera:
Thanks for the clarification, @dscerutti!

dscerutti:
The additional validation I am doing is a control experiment of sorts--I have already verified that the energies of systems created according to each version of parm15ipq_10.3.dat are consistent when mdgx processes the results, and that both parameter sets cover all details of every amino acid in mdgx's view.  The program's force field fitting tools automatically read in topologies (created by tleap) and cross-reference every unique parameter that they find with one in the parm.dat file also referenced by the run.  The program needs to find complete correspondence EXCEPT for the phase amplitude--so atom types, phase angles, and periodicities need to match (the amplitudes are free because it's assumed that that's what is being fitted while the other details are fixed until someone invents entirely new parameters rather than just retuning what is there.) The fact that I can remove the original wildcard dihedrals completely and replace them with the specifics that I did, and not have mdgx complain "you didn't tell me where this parameter came from" is solid evidence that I am covering all of the bases that I need to.  Once it's found all of the correspondence, mdgx takes the parameters from the parm.dat file and applies them to compute energies for each structure (prior to optimizing the amplitudes).   The fact that mdgx computes consistent energies for the old and new versions of the file is then solid proof that I have all of the right parameters and the correct amplitudes for each of them.  I don't expect you all to take so much interest, but in all that I can now see how the idiosyncracy of tleap slipped through my net and allowed this problem in the first place.  Apart form that, I still get some satisfaction out of the way mdgx checks everything for consistency and shows you what goes where.  It was with those tools that Karl first noticed something amiss with the histidine dihedrals, so I am sorry that in my haste in late 2015 I said "don't worry about it."

Now, I've already got good reason to think that the file I sent you is correct, but the control experiment I wanted to do is that I should see energy differences when I use SANDER to run single point calculations on topologies created by tleap (tleap always creates the topologies), as opposed to single point calculations made by the mdgx force field tools (which, as I explained, take the same topologies but reference parameter values all the way back to the parm.dat).  Indeed, I see those energy differences if, and only if, the molecules contain HIE, HID, or HIP, so the parameters were confined to histidines and the replacements I have introduced have not been applied anywhere unexpected.

What I was asking earlier is that Rafal should not only find that the file converts fine, but that it passes the test, vis-à-vis consistency with OpenMM and tleap-created topologies, which the previous file failed.  However, I've done enough to convince myself that the new file is correct so, Rafal, at your leisure... which I expect you have precious little of given the pending OpenMM release.

Finally... I ran some dynamics on ubiquitin, where there is a solvent-exposed histidine, and after 32ns the ring structure appears... flat and healthy.  Take it away, I'm patching Amber.

rafwiewiora:
@dscerutti by 'converts fine' I meant it converts and passes our test of energies being the same to `1e-5` relative error between tleap- and OpenMM-built systems. You can see the energies from both and relative errors in the log: https://github.com/choderalab/openmm-forcefields/blob/5dd3e3bf4052e79bd9570824b8a1194eca0ebe05/amber/log.csv - we're testing on Ala-Ala-Ala and villin headpiece.

dscerutti:
Great to hear!  I'll document that as I said.

dscerutti:
Patch is active in Amber.  Happy Thanksgiving!

jchodera:
Huzzah! Thanks!

PR #51, opened 2017-11-20
Add prefixes for protein, DNA, and RNA leaprc conversions

jchodera:
Addresses https://github.com/choderalab/openmm-forcefields/issues/46 by adding `protein-`, `DNA-`, and `RNA-` prefixes for leaprc conversions.


jchodera:
This seems to be working. (Builds are timing out as expected.)

PR #58, opened 2018-01-23
Update CHARMM three-site water conversion

jchodera:
- [x] Add capability to specify residues to be eliminated from CHARMM parameter files during conversion
- [x] Regenerate waters_ions_default.xml with TP3M template culled
- [x] Add ability to test parameterization on test systems.
- [x] Test all three-site waters with `ForceField.createSystem()`
- [x] Add energy comparison between ParmEd CHARMM reader and OpenMM ffxml route for three-site waters


jchodera:
@peastman : This PR contains updates to all the three-site water `xml` files. You should be able to update those now. I've also tested the energies of a waterbox against the ParmEd CHARMM reader.

The four-site and five-site waters will be handled in a separate PR.

Note that this PR requires a modified version of ParmEd: https://github.com/ParmEd/ParmEd/pull/942

jchodera:
@peastman: Here are convenient links to the updates three-site water files:
```
* [waters_ions_default.xml](https://raw.githubusercontent.com/choderalab/openmm-forcefields/0905e8cd49455b7d3690aeb4689ca7a31188983c/charmm/ffxml/waters_ions_default.xml)
* [waters_ions_spc.xml](https://raw.githubusercontent.com/choderalab/openmm-forcefields/0905e8cd49455b7d3690aeb4689ca7a31188983c/charmm/ffxml/waters_ions_spc.xml)
* [waters_ions_spc_e.xml](https://raw.githubusercontent.com/choderalab/openmm-forcefields/0905e8cd49455b7d3690aeb4689ca7a31188983c/charmm/ffxml/waters_ions_spc_e.xml)
* [waters_ions_tip3p_ewald_b.xml](https://raw.githubusercontent.com/choderalab/openmm-forcefields/0905e8cd49455b7d3690aeb4689ca7a31188983c/charmm/ffxml/waters_ions_tip3p_ewald_b.xml)
* [waters_ions_tip3p_ewald_f.xml](https://raw.githubusercontent.com/choderalab/openmm-forcefields/0905e8cd49455b7d3690aeb4689ca7a31188983c/charmm/ffxml/waters_ions_tip3p_ewald_f.xml)

**Comment by peastman** [comment]
If you're having problems getting ParmEd to process the four and five point models, we could just delete the extra line by hand.

**Comment by jchodera** [comment]
I just need to add line pair conversion support.

I'd rather not do this by hand, but you are welcome to contribute a ParmEd PR to add that feature if you have time! I just have one more thing to clean up in my previous PR tonight, but can hopefully take a stab at the EPs tomorrow.

**Comment by jchodera** [comment]
The changes for three-site waters have been merged into ParmEd (https://github.com/ParmEd/ParmEd/pull/942). I'm working on the EPs now.

**Comment by jchodera** [comment]
@peastman : I've updated the multisite water models with my preliminary implementation of your [conversion scheme](https://github.com/pandegroup/openmm/blob/master/devtools/forcefield-scripts/processCharmmForceField.py#L320-L333):
* [`waters_ions_tip4p.xml`](https://github.com/choderalab/openmm-forcefields/blob/fix-duplicate-waters/charmm/ffxml/waters_ions_tip4p.xml)
* [`waters_ions_tip4p_ew.xml`](https://github.com/choderalab/openmm-forcefields/blob/fix-duplicate-waters/charmm/ffxml/waters_ions_tip4p_ew.xml)
* [`waters_ions_tip4p_2005.xml`](https://github.com/choderalab/openmm-forcefields/blob/fix-duplicate-waters/charmm/ffxml/waters_ions_tip4p_2005.xml)
* [`waters_ions_tip5p.xml`](https://github.com/choderalab/openmm-forcefields/blob/fix-duplicate-waters/charmm/ffxml/waters_ions_tip5p.xml)
* [`waters_ions_tip5p_ew.xml`](https://github.com/choderalab/openmm-forcefields/blob/fix-duplicate-waters/charmm/ffxml/waters_ions_tip5p_ew.xml)



I note that the `<VirtualSite>` definitions in my conversion differ from your [TIP5P implementation](https://github.com/pandegroup/openmm/blob/f7935fef719ed0cd8292afef676c39d881b6a960/wrappers/python/simtk/openmm/app/data/tip5p.xml#L14-L15). Do these look equivalent to you?

Also, I would love to test these against PDB/PSF file for multisite water boxes, but I'm not entirely sure how to generate pairs of PDB and PSF files for multisite waters. I realize that I'm not even sure of how residue templates containing virtual sites are matched within `ForceField` to assign nonbonded parameters---does this use virtual site names, or atom ordering, or something else?

**Comment by peastman** [comment]
> I note that the <VirtualSite> definitions in my conversion differ from your TIP5P implementation. Do these look equivalent to you?

The existing ones use OutOfPlaneSite rather than LocalCoordinatesSite.  I can't tell from looking at them whether they're equivalent, since they each involve a bit of computation.  We just need to check whether the virtual sites come out at the same locations.

> I'm not entirely sure how to generate pairs of PDB and PSF files for multisite waters.

Creating a PDB file is easy: use `Modeller.addExtraParticles()` to add the virtual sites, then write it out to a PDB file.  I don't know how PSF files are created, but I assume CHARMM includes a tool for doing it?

> I'm not even sure of how residue templates containing virtual sites are matched within ForceField

Extra particles are identified by the fact that their `element` field is set to `None` (both in the residue and in the template).  They get matched just like any other particle: element must be the same and set of bonds must be the same.  It does add a special case for extra particles that it will always prefer one whose name exactly matches over one that doesn't.

**Comment by jchodera** [comment]
> Extra particles are identified by the fact that their element field is set to None (both in the residue and in the template). They get matched just like any other particle: element must be the same and set of bonds must be the same. It does add a special case for extra particles that it will always prefer one whose name exactly matches over one that doesn't.

What do "bonds" mean for virtual sites? They aren't actually chemically bonded to anything. Are the atoms used to define their positions treated as bonds by `ForceField` or `Topology`?

Thankfully, in the water models here, any virtual sites are always equivalent.

**Comment by peastman** [comment]
In principle a virtual site can have bonds, but in this case they don't.  If the atom names in the template exactly match the ones in the residue, it will use them to do the assignment.  Otherwise, it's arbitrary which way they'll get assigned.  But since the two sites are identical, it doesn't matter which way they get assigned.

**Comment by jchodera** [comment]
> In principle a virtual site can have bonds, but in this case they don't.

You mean, they *shouldn't* in this case. They currently do, because that's what CHARMM specifies, but I should also remove them because OpenMM insists water residues should not have anything but H-O bonds. Is that right?

**Comment by jchodera** [comment]
@peastman : I'm having some trouble here. The new [TIP4P ffxml files](https://github.com/choderalab/openmm-forcefields/blob/fix-duplicate-waters/charmm/ffxml/waters_ions_tip4p.xml#L28-L36) no longer have bonds to virtual sites, but it looks like [my test](https://github.com/choderalab/openmm-forcefields/blob/fix-duplicate-waters/charmm/test_charmm.py#L55) is failing because it thinks there are angles between virtual sites and real atoms:
```
Testing waterbox TIP4P
  File "/Users/choderaj/miniconda3/lib/python3.6/site-packages/simtk/openmm/app/charmmpsffile.py", line 519, in loadParameters
    ang.angle_type = parmset.angle_types[key]
KeyError: ('HT', 'OT', 'LP')

During handling of the above exception, another exception occurred:

  File "test_charmm.py", line 249, in <module>
    main()
  File "test_charmm.py", line 34, in main
    test_charmm()
  File "test_charmm.py", line 61, in test_charmm
  File "test_charmm.py", line 100, in compare_energies
    openmm_system = openmm_psf.createSystem(openmm_toppar, **system_kwargs)
  File "/Users/choderaj/miniconda3/lib/python3.6/site-packages/simtk/openmm/app/charmmpsffile.py", line 743, in createSystem
    self.loadParameters(params.condense())
  File "/Users/choderaj/miniconda3/lib/python3.6/site-packages/simtk/openmm/app/charmmpsffile.py", line 525, in loadParameters
    raise MissingParameter('Missing angle type for %r' % ang)
simtk.openmm.app.internal.charmm.exceptions.MissingParameter: Missing angle type for <Angle; <Atom 1; 1 TIP4 [H1: HT]>-<Atom 0; 1 TIP4 [OH2: OT]>-<Atom 3; 1 TIP4 [OM: LP]>; type=None>
```
All the files need to reproduce are checked into this PR:
```
cd charmm
python test_charmm.py 
```

**Comment by peastman** [comment]
> but I should also remove them because OpenMM insists water residues should not have anything but H-O bonds. Is that right?

Correct.

**Comment by peastman** [comment]
I assume that's also why you're getting the error about the missing angle.  There isn't actually a HT-OT-LP angle, but it thinks there is because it thinks there's a OT-LP bond.

**Comment by jchodera** [comment]
But where is the bond defined in the new ffxml file?
https://github.com/choderalab/openmm-forcefields/blob/fix-duplicate-waters/charmm/ffxml/waters_ions_tip4p.xml#L28-L36

Unless I'm going blind, my updated ParmEd branch removes bonds to the lone pair.

**Comment by peastman** [comment]
The error message is coming from CharmmPsfFile, not from the ForceField.

**Comment by jchodera** [comment]
cc: https://github.com/ParmEd/ParmEd/pull/945

**Comment by jchodera** [comment]
> The error message is coming from CharmmPsfFile, not from the ForceField.

Ah! Thanks @peastman. I had found these TIP4P input files prepared by a different group, but I suppose I should just figure out how to create my own CHARMM PSF files for multi-site waters for testing. (Sadly, it doesn't look like CHARMM-GUI does this.)

**Comment by jchodera** [comment]
@peastman : I couldn't get any of the CHARMM- or VMD-based tools to work for generating PSF files for multisite water boxes, so I used OpenMM's `Modeller` to add the extra sites and `ParmEd` to somewhat-manually create corresponding PSF files. These are now checked into the repo.

I'm now getting an OpenMM error that seems to suggest that the OpenMM PSF file reader is trying to look up the atomic number of the lone pair virtual sites:
```
Testing waterbox TIP4P
  File "test_charmm.py", line 253, in <module>
    main()
  File "test_charmm.py", line 34, in main
    test_charmm()
  File "test_charmm.py", line 65, in test_charmm
  File "test_charmm.py", line 104, in compare_energies
    openmm_system = openmm_psf.createSystem(openmm_toppar, **system_kwargs)
  File "/Users/choderaj/miniconda/lib/python3.6/site-packages/simtk/openmm/app/charmmpsffile.py", line 744, in createSystem
    hasbox = self.topology.getUnitCellDimensions() is not None
  File "/Users/choderaj/miniconda/lib/python3.6/site-packages/simtk/openmm/app/charmmpsffile.py", line 648, in topology
    elem = element.Element.getByAtomicNumber(atomic_num)
  File "/Users/choderaj/miniconda/lib/python3.6/site-packages/simtk/openmm/app/element.py", line 107, in getByAtomicNumber
    return Element._elements_by_atomic_number[atomic_number]
KeyError: 0
```
Presumably, this means that my PSF file is not correctly specifying the lone pairs, but I'm not entirely sure whether ParmEd or OpenMM even support writing or reading of lone pairs correctly. For example, [this section](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/charmmpsffile.py#L334-L347) suggests that lone pairs are not implemented.

**Comment by peastman** [comment]
It's possible CharmmPsfFile doesn't handle virtual sites correctly.  Maybe @swails can comment on that, since he wrote them?  Also, Alex MacKerell's group just submitted a paper on using the Drude force field with OpenMM.  They had to make some changes to the CHARMM reading classes to fully support that, which I think may have involved changes to virtual site support.  They should be sending a PR with the changes fairly soon.

Here's a simple script I wrote for doing a sanity check of the water models.  For each model that has multiple supposedly identical implementations, it computes the energy of a water box so we can make sure they match.  Though note that the CHARMM versions of TIP3P and SPC/E aren't really identical to the others, since they add a Lennard-Jones term to the hydrogens, so those aren't expected to match exactly.

```python
from simtk.openmm import *
from simtk.openmm.app import *

tests = [
    ['tip3p.pdb', ['tip3p.xml', 'amber14/tip3p.xml', 'charmm36/water.xml']],
    ['tip3p.pdb', ['tip3pfb.xml', 'amber14/tip3pfb.xml']],
    ['spce.pdb', ['spce.xml', 'amber14/spce.xml', 'charmm36/spce.xml']],
    ['tip4pew.pdb', ['tip4pew.xml', 'amber14/tip4pew.xml', 'charmm36/tip4pew.xml']],
    ['tip4pew.pdb', ['tip4pfb.xml', 'amber14/tip4pfb.xml']],
    ['tip4p.pdb', ['tip5p.xml', 'amber14/tip5p.xml', 'charmm36/tip5p.xml']]
]

for test in tests:
    pdb = PDBFile(forcefield._dataDirectories[0]+'/'+test[0])
    for xml in test[1]:
        ff = ForceField(xml)
        system = ff.createSystem(pdb.topology, nonbondedMethod=PME)
        integrator = VerletIntegrator(0.001)
        context = Context(system, integrator)
        context.setPositions(pdb.positions)
        context.applyConstraints(1e-6)
        print(xml, context.getState(getEnergy=True).getPotentialEnergy())
    print()
```

jchodera:
The goal here is to ensure we've converted the parameters (and all accompanying infrastructure for virtual sites) correctly, so we somehow need to compare energies with either CHARMM directly or a validated implementation of the CHARMM parameter file readers. Ultimately, we'll need to do one of those things if we want to claim these models are correct.

We can update all of the CHARMM water models in OpenMM now (since we think they should at least *work* now), but continue to pursue energy comparisons in parallel.

Would you be able to focus on bringing the CHARMM readers in ParmEd and OpenMM up to the point where they can read/write lone pairs in PSF files, while I focus on validating directly against CHARMM energies directly by extending the [OpenMMEnergyComparisons](https://github.com/choderalab/OpenMMEnergyComparisons) we used for the OpenMM 7 paper?

peastman:
I think we should compare to energies computed by CHARMM.  If the PSF readers don't currently support virtual sites, adding that could be a big project, and then we'd need to validate that it worked correctly... by comparing to CHARMM.  So that comparison is necessary either way.

If the XML files linked above are now correct as far as we know (all extra bonds removed), I can go ahead and add them to OpenMM and we can get started building the new release candidate.

jchodera:
> I think we should compare to energies computed by CHARMM. If the PSF readers don't currently support virtual sites, adding that could be a big project, and then we'd need to validate that it worked correctly... by comparing to CHARMM. So that comparison is necessary either way.

OK, can you spearhead that project (likely for post-7.2)? We probably want to make a note of the current limitations in the user guide and API docs for now.

I'll focus on getting the direct-to-CHARMM comparisons working.

> If the XML files linked above are now correct as far as we know (all extra bonds removed), I can go ahead and add them to OpenMM and we can get started building the new release candidate.

I'd double-check the ffxml files yourself, but they are "correct" as far as I can am aware. However, the VirtualSite(s) are complex enough to not be able to easily check those by eye.

peastman:
Got it.  The script I posted above should check the virtual sites sufficiently, since we expect the CHARMM versions of TIP4P-Ew and TIP5P to exactly match the others.  Any error in the virtual site definitions would change the energy.  I'll check that.

jchodera:
I'll go ahead and merge this for now.

peastman:
It's still failing, because apparently ForceField is insisting on an exact name match for the extra particles.  It really shouldn't require that, since in this case there's no ambiguity.  I'll make the code change to fix that (fortunately only a very small one).

With all the latest files, here's the output of the script above:

```
tip3p.xml -36024.58632622575 kJ/mol
amber14/tip3p.xml -36024.58632622575 kJ/mol
charmm36/water.xml -36867.13376627362 kJ/mol

tip3pfb.xml -40807.8225592541 kJ/mol
amber14/tip3pfb.xml -40807.8381842541 kJ/mol

spce.xml -41680.63535323567 kJ/mol
amber14/spce.xml -41680.61972823567 kJ/mol
charmm36/spce.xml -42972.27446182314 kJ/mol

tip4pew.xml -42067.740648921404 kJ/mol
amber14/tip4pew.xml -42067.771898921404 kJ/mol
charmm36/tip4pew.xml -42071.37879702408 kJ/mol

tip4pfb.xml -44439.68939898891 kJ/mol
amber14/tip4pfb.xml -44439.67377398891 kJ/mol

tip5p.xml 1630.620064353665 kJ/mol
charmm36/tip5p.xml 1603.0475411914886 kJ/mol
```

Agreement between the old water models and the Amber ones is excellent.  The CHARMM ones are close, but not exact.  That isn't surprising for TIP3P and SPC/E, since they have the extra LJ sites, but I'd expect the others to be closer than they are.  I'll investigate.

jchodera:
Thanks!

peastman:
The difference comes from very slight differences in how the LJ parameters are defined.  In the case of TIP4P-Ew, for example, the standard and Amber versions both give sigma as 0.316435, but the CHARMM version gives it as 0.31642940670908576.  Epsilon and charge are identical in all versions.  The positions of the virtual sites are also identical.

PR #59, opened 2018-01-28
Update CHARMM conversion and add protein:ligand tests

jchodera:
- [x] Disable `osx` tests since LEaP segfaults during some conversions
- [x] Disable CHARMM multisite water tests since OpenMM PSF reader can't handle these
- [x] Update CHARMM parameter files to fix `CTER` omitted `<RemoveExternalBond>` tag
- [x] Update CHARMM protein:ligand tests
- [ ] Fix CHARMM protein:ligand ions for tests


jchodera:
@peastman : I just found another issue with `CTER` patches omitting the `<RemoveExternalBond>` tag:
```
    <Patch name="CTER">
      <ChangeAtom charge="0.34" name="C" type="CC"/>
      <AddAtom charge="-0.67" name="OT1" type="OC"/>
      <AddAtom charge="-0.67" name="OT2" type="OC"/>
      <RemoveAtom name="O"/>
      <AddBond atomName1="C" atomName2="OT2"/>
      <AddBond atomName1="C" atomName2="OT1"/>
      <RemoveBond atomName1="C" atomName2="O"/>
    </Patch>
```
Trying to figure out why this is happening, since `<RemoveExternalBond>` is correctly added for `NTER`.

jchodera:
@peastman : I've updated the `charmm36_nowaters.xml` to fix some issues with the `CTER` patch missing a `<RemoveExternalBond/>` tag. 

I'm now running into issues with testing the protein:solvent systems I had added because of unrecognized ions, presumably because the `POT` atoms aren't recognized as being potassium:
```
Testing 1VII solvated
/Users/choderaj/miniconda/lib/python3.6/site-packages/simtk/openmm/app/charmmpsffile.py:341: CharmmPSFWarning: Detected PSF molecule section that is WRONG. Resetting molecularity.
  'Resetting molecularity.', CharmmPSFWarning)
/Users/choderaj/miniconda/lib/python3.6/site-packages/ParmEd-3.0.0+20.g84a71952.dirty-py3.6-macosx-10.9-x86_64.egg/parmed/charmm/psf.py:320: CharmmWarning: Detected PSF molecule section that is WRONG. Resetting molecularity.
  'Resetting molecularity.', CharmmWarning)
/Users/choderaj/miniconda/lib/python3.6/site-packages/ParmEd-3.0.0+20.g84a71952.dirty-py3.6-macosx-10.9-x86_64.egg/parmed/charmm/psf.py:651: ParameterWarning: ('CC', 'OC', 'X', 'X') and ('CC', 'CT2', 'X', 'X') match improper ('CC', 'CT2', 'OC', 'OC'). Using ('CC', 'OC', 'X', 'X')
  altkey1, key), ParameterWarning)
/Users/choderaj/miniconda/lib/python3.6/site-packages/ParmEd-3.0.0+20.g84a71952.dirty-py3.6-macosx-10.9-x86_64.egg/parmed/charmm/psf.py:651: ParameterWarning: ('CC', 'OC', 'X', 'X') and ('CC', 'CT2', 'X', 'X') match improper ('CC', 'CT2', 'OC', 'OC'). Using ('CC', 'OC', 'X', 'X')
  altkey1, key), ParameterWarning)
/Users/choderaj/miniconda/lib/python3.6/site-packages/ParmEd-3.0.0+20.g84a71952.dirty-py3.6-macosx-10.9-x86_64.egg/parmed/charmm/psf.py:651: ParameterWarning: ('CC', 'OC', 'X', 'X') and ('CC', 'CT1', 'X', 'X') match improper ('CC', 'CT1', 'OC', 'OC'). Using ('CC', 'OC', 'X', 'X')
  altkey1, key), ParameterWarning)
  File "test_charmm.py", line 258, in <module>
    main()
  File "test_charmm.py", line 34, in main
    test_charmm()
  File "test_charmm.py", line 70, in test_charmm
  File "test_charmm.py", line 131, in compare_energies
    system_openmm = ff.createSystem(pdbfile.topology, **system_kwargs)
  File "/Users/choderaj/miniconda/lib/python3.6/site-packages/simtk/openmm/app/forcefield.py", line 1128, in createSystem
    raise ValueError('No template found for residue %d (%s).  %s' % (res.index+1, res.name, _findMatchErrors(self, res)))
ValueError: No template found for residue 4467 (POT).  This might mean your input topology is missing some atoms or bonds, or possibly that you are using the wrong force field.
```
I had prepared these files in CHARMM-GUI. Presumably, `POT` and `CLA` are not PDB standards, so there's no simple way to detect the element types produced by CHARMM-GUI to read these into OpenMM?

This is going to make testing in OpenMM and CHARMM difficult.

peastman:
As long as the element is specified correctly both in the XML file and the Topology, it shouldn't care.  Which one are they not specified in?

jchodera:
The [PDB file](https://github.com/choderalab/openmm-forcefields/blob/1b803c5795dc0efb90d01b713b9f41b06de8a1ef/charmm/tests/charmm-gui/1VII/step2_solvator.pdb) has entries like
```
ATOM  13887  POT POT     1      -3.560 -23.634  25.385  1.00  0.00      POT 
...
ATOM  13899  CLA CLA     1     -10.915  10.080  24.165  1.00  0.00      CLA 
```
so I imagine OpenMM's `PDBFile` isn't picking up on the correct elements for these atoms.

jchodera:
Yep---it thinks they are phosphorous and chlorine:
```
lski2414:charmm choderaj$ python
fPython 3.6.3 | packaged by conda-forge | (default, Nov  4 2017, 10:13:32) 
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from simtk.openmm import app
>>> pdbfile = app.PDBFile('tests/charmm-gui/1VII/step2_solvator.pdb')
>>> atoms = [atom for atom in pdbfile.topology.atoms()]
>>> atoms[-20].element
<Element phosphorus>
>>> atoms[-10].element
<Element chlorine>
```

peastman:
Those strings "POT" and "CLA" at the ends of the lines don't belong there.  They're in columns 73-75, which don't have any defined meaning (see http://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#ATOM).  The element symbol should be in columns 77-78.  If you add the correct symbols in those columns, it should work correctly.

jchodera:
> Those strings "POT" and "CLA" at the ends of the lines don't belong there. They're in columns 73-75, which don't have any defined meaning (see http://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#ATOM). The element symbol should be in columns 77-78. If you add the correct symbols in those columns, it should work correctly.

The `ATOM` records are also supposed to be only used for "standard amino acids and nucleotides", so this is a gross violation of the PDB standard in multiple ways. But these files are generated by CHARMM-GUI. Do we really want to prevent people from using PDB files generated with CHARMM-GUI?

To carry out the validation, it looks like I'd have to prepare PDB files that CHARMM can read alongside matching PDB files that OpenMM can read. I could write a script to do the translation, but at that point, why not add a mode to `PDBFile` to accommodate CHARMM-GUI PDB files?

peastman:
Then let's contact the CHARMM-GUI developers and ask them to fix it so it generates valid PDB files.  There just isn't anything we can do about it on our end.  How are we supposed to know that an atom with a nonstandard name beginning with P and nothing in the element symbol field is a potassium?

jchodera:
> Then let's contact the CHARMM-GUI developers and ask them to fix it so it generates valid PDB files.

I've emailed Wonpil Im (who receives emails directed to the CHARMM-GUI email form) and CC'd you.

peastman:
Great!  It sounds like we can move ahead with the new release candidate.  Where can I get the updated charmm36_nowaters.xml?

jchodera:
@peastman : All the latest CHARMM `ffxml` files are now in `master` in [this directory](https://github.com/choderalab/openmm-forcefields/tree/master/charmm/ffxml).

I'm a bit worried about that discrepancy, but this set of `ffxml` files is definitely a big improvement on the last. We can figure out what the remaining issues are in parallel with more testing.

PR #64, opened 2018-04-24
Update CHARMM ffxml by culling duplicate patches

jchodera:
This PR addresses the ffxml CHARMM patch duplication issue discovered in https://github.com/pandegroup/openmm/issues/2047

The updated [`charmm36_nowaters.xml`](https://github.com/choderalab/openmm-forcefields/raw/unique-patches/charmm/ffxml/charmm36_nowaters.xml) removes duplicate patches.

@peastman: You should be able to download this version and update the OpenMM version.

This PR requires ParmEd fix https://github.com/ParmEd/ParmEd/pull/978


jchodera:
I've regenerated the CHARMM ffxml parameters using https://github.com/ParmEd/ParmEd/pull/1001

Note that this leads to slightly different parameters from before, but my understanding is that this is because the logic used to expand generics does not result in unique matches.

peastman:
Thanks!  Is this ready to merge into OpenMM?

peastman:
Ok, I'll post a PR with these changes.

PR #78, opened 2019-10-27
Update to MolSSI cookiecutter

jchodera:


jchodera:
We're now up to speed with the latest MolSSI Cookiecutter!

There are still some issues with the CHARMM conversion tests, but I suspect this is due to using an old ParmEd. We'll see what happens when we update to AmberTools 19.

PR #79, opened 2019-10-27
Update Amber conversion to AmberTools 19.10 and add GAFF 1/2 residue template generators

jchodera:
This PR updates the Amber conversion to AmberTools 18 (in preparation for an update to AmberTools 19.6 with ff19SB support!).

In addition, we split out, version, and convert all the historical GAFFs I could find, which includes 1.4, 1.8, 1.81, 2.1, and 2.11. We'll hopefully get the complete set from Junmei Wang after GAFF 2 is officially released.


jchodera:
Woohoo!

PR #95, opened 2020-01-22
Fix GAFF AM1-BCC charging failures

jchodera:
This PR fixes charging failures that occur when some ligands are parameterized with the `GAFFTemplateGenerator`. 

cc: https://github.com/openforcefield/openforcefield/issues/492


j-wags:
Does openmm-forcefields have the same philosophy as SMIRNOFF about generating its own input conformers for charge calculation? If so, then this should resolve the issue.

jchodera:
> Does openmm-forcefields have the same philosophy as SMIRNOFF about generating its own input conformers for charge calculation? If so, then this should resolve the issue.

Yes, that's the idea---to attempt to use AM1-BCC ELF10 best practices (when available).

hannahbrucemacdonald:
Please can I have permissions for this repo!

jchodera:
@hannahbrucemacdonald : I'll let the tests run and merge in the morning!

jchodera:
Hooray! Test pass!

PR #96, opened 2020-01-23
Changes made to `fix-gaff`

hannahbrucemacdonald:
I've removed the input files for the systems that I haven't finished setting up yet, as I thought that was less confusing.

Added how the files were generated to the README in the directory

Added details to the changelog in the main README


jchodera:
Thanks!

Issue #99, opened 2020-02-03
Harmonize prefixes for AMBER protein force field additions


jchodera:
Currently, we have an issue where some converted AMBER protein force field files contain the prefixe `protein-` while others do not, causing seemingly straightforward operations like bringing in the phosphoaminoacid templates:
```python
from simtk.openmm.app import ForceField
ff = ForceField('amber/protein.ff14SB.xml', 'amber/phosaa10.xml')
```
to fail:
```
Traceback (most recent call last):
 File “<stdin>“, line 1, in <module>
 File “/home/brucemah/miniconda3/envs/omm_source/lib/python3.6/site-packages/simtk/openmm/app/forcefield.py”, line 185, in __init__
  self.loadFile(files)
 File “/home/brucemah/miniconda3/envs/omm_source/lib/python3.6/site-packages/simtk/openmm/app/forcefield.py”, line 278, in loadFile
  template.atoms.append(ForceField._TemplateAtomData(atomName, typeName, self._atomTypes[typeName].element, params))
KeyError: ‘N’
```
We therefore need to decide whether we want to harmonize these by prefixing _all_ protein atom types with `protein-` so these will be compatible.

We can support this by adding [a line to the YAML](https://github.com/openmm/openmmforcefields/blob/master/amber/biopolymer.yaml#L9):
```
  Prefix: protein
```
for each file we want to convert and prepend prefixes for atom types.

Questions:
* Which files should we do this with?
* Should we also do this with all files in `oldff/`?
* How should we augment the tests to catch these incompatibilities

cc: @rafwiewiora @hannahbrucemacdonald @peastman 

We should likely include the fixed parameters and information on how to use the small molecule parameters in an upcoming OpenMM release.

hannahbrucemacdonald:
Is there something inherently different between those that are prefixed, and those that aren't?

jchodera:
From what I recall, @peastman was concerned about atom type and class name collisions between different ffxml files packaged with OpenMM. I believe that's why we added prefixes for those ffxml files that were to be packaged with OpenMM, but we likely need to do this for all recipes if we are going to do this for any.

jchodera:
@peastman: Some of the AMBER force field files we are converting contain only protein, only lipid, or only nucleic acid parameters, but others (such as `leaprc.PM3`) contain a mixture of multiple kinds. 

[...]

PR #100, opened 2020-02-07
[Do not merge] adding `protein-` prefix to phosaa10.xml

hannahbrucemacdonald:
Adding `protein-` prefix to the file, so as to be consistent with atom namings in the protein.xxx.xml files in the repo.

This is not fixed - some terms still seem to be missing

Associated to Issue #99


jchodera:
Closing this out-of-date PR.

PR #101, opened 2020-02-11
[WIP] Adding all JACS systems to the repo

hannahbrucemacdonald:
WARNING: currently CDK2 has CONECT lines in the `protein.pdb` while the other systems do not. While openmm can handle this, it would be better for all the protein files to be consistent


hannahbrucemacdonald:
pinging @jchodera

jchodera:
Thanks!

PR #109, opened 2020-04-08
Throw exception if nonbondedMethod provided to forcefield_kwargs in SystemGenerator; use OpenMM 7.4.2 rc

### LINKED ISSUE #108: Throw exception if `nonbondedMethod` is fed to SystemGenerator forcefield_kwargs
cc: https://github.com/choderalab/perses/issues/661

We should throw an informative exception if `nonbondedMethod` is fed to `SystemGenerator` `forcefield_kwargs`, informing that these should go into either `periodic_forcefield_kwargs` or `nonbonded_forcefield_kwargs`.

jchodera:
Fixes #108

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,6 +62,7 @@ script:
   #- pytest -v --cov=openmmforcefields openmmforcefields/tests
   - pytest -v -x -s --log-cli-level INFO --cov=openmmforcefields openmmforcefields/tests/test_amber_import.py
   - pytest -v -x -s --log-cli-level INFO --cov=openmmforcefields openmmforcefields/tests/test_template_generators.py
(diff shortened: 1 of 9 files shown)


hannahbrucemacdonald:
`Miniconda MD5 mismatch`


This seems to be a conda version issue. Maybe we should be pinning conda? No idea why it passes on master and not here. @jaimergp do you have any idea?

jaimergp:
That can happen when the Anaconda guys update the Miniconda installer and then take a bit to update the md5 hash on the website. A retry within the day should work... That said, this could be migrated to GH Actions to avoid the manual Miniconda installation and other perks :) Would you like some help there? We did it for perses and openforcefield, so the process is now very straightforward.

hannahbrucemacdonald:
Thanks! I will retrigger the build tomorrow to check the tests.

Wrt to GH Actions - up to you + John!

PR #117, opened 2020-06-05
Added 2019 CHARMM Drude force field

peastman:
This requires the changes from https://github.com/ParmEd/ParmEd/pull/1091, which will hopefully be merged soon.

For the moment I've omitted the nucleic acids and carbohydrates.  If you try to convert them, ParmEd throws an exception because they involve situations it can't handle.  Eventually it would be nice to add them, but that will require further changes to ParmEd, and possibly to OpenMM.

I've tested this locally on one system to make sure it agrees with CHARMM.  I haven't tried to hook it into the testing system in this repository.  The Drude forcefield needs to be handled a bit differently from other forcefields.


jchodera:
Thanks, @peastman! 

Are you including the resulting force fields directly in `openmm`, or will `openmmforcefields` be providing them to end users? 

If the former, it would be great if we can get a new `parmed` release cut after https://github.com/ParmEd/ParmEd/pull/1091 is merged, and then a patch release out to `ambertools` before we merge this (since we test against and use as a dependency the conda-forge `ambertools`).

If the latter, we can merge and cut a release right away and then work on getting the parmed updates out.

peastman:
I'll be including it directly in OpenMM.

jchodera:
OK! Can you lobby for the ParmEd release and inclusion in ambertools after your PR is merged?

andysim:
The latest CHARMM FF release from Alex's website (toppar_c36_jul19) also [updates the non-polarized protein force field](https://www.nature.com/articles/nmeth.4067) a little: the pertinent file is `par_all36m_prot.prm`.  Are there any plans to include that xml also?  I can easily generate it myself, but don't want to duplicate efforts.

peastman:
I think there's still a bug somewhere.  The agreement with CHARMM is pretty good, but not quite as precise as it ought to be.  I'm investigating.

peastman:
Actually, I think the error may be in the CHARMM numbers.  I'm combining the most recent toppar_drude_master_protein_2019g.str with a PSF that was generated for the earlier toppar_drude_master_protein_2013f.str version of the force field, so the charges are wrong.  CHARMM-GUI doesn't seem to have added support for the latest version yet.  Can anyone tell me how to create a PSF manually that matches the new version?

andysim:
If you have the tarball from CHARMM-GUI, I can modify that using CHARMM itself (hopefully)

peastman:
Thanks so much!

https://send.firefox.com/download/5f50b09329c1e3fc/#tC9iRQAJbwXDgLV2Aadolg

andysim:
I am clueless when it comes to CHARMM, but maybe this is correct?

https://send.firefox.com/download/59958c941de0a9cf/#R1Mz5MM3-C99V-xP2qbLWA

andysim:
I'm happy to try again if it isn't.  I just switched out the files listed in the stream file and then ran the step1 and step2 files through vanilla CHARMM.  That _should_ be all that's needed.

peastman:
That looks perfect!  The charges listed in the PSF now match what's in the ParmEd-converted force field, and the energy agreement is much better.

My own knowledge of how to use CHARMM doesn't go much further than, "Go to CHARMM-GUI and hope it has a command to do what I want." :)

andysim:
For the record, the reciprocal space energy of `71.98589` becomes `71.98622` if I crank the grid up to 64x64x64.

jchodera:
@peastman: Do you know if the parmed changes have been released into an ambertools 20 patch release yet?

peastman:
It looks like the [latest ambertools release](https://anaconda.org/conda-forge/ambertools/files) was posted eight days ago.  I don't know whether it includes those changes.

peastman:
I notice this has never been merged.  Is there anything else you need from me?

jchodera:
@peastman : Whoops, can you just rebase or update to the latest master and then we can merge?

peastman:
I updated the branch.  The CI builds are failing to start up properly:

```
$ openssl aes-256-cbc -K $encrypted_c0daf35c25ce_key -iv $encrypted_c0daf35c25ce_iv -in devtools/travis-ci/oe_license.txt.enc -out devtools/travis-ci/oe_license.txt -d
iv undefined
The command "openssl aes-256-cbc -K $encrypted_c0daf35c25ce_key -iv $encrypted_c0daf35c25ce_iv -in devtools/travis-ci/oe_license.txt.enc -out devtools/travis-ci/oe_license.txt -d" failed and exited with 1 during .
```

jchodera:
That's an expired OpenEye license. I'll merge and fix!

jchodera:
@peastman : The issue was not the license, but opening a PR from your fork. If you open it from a branch of the main repo, the Travis trusted secrets will allow the license to be decrypted and the tests to run.

I'll see if I can move the OpenEye license to be optional since the OpenFF toolkit should be able to run with just free tools instead.

PR #124, opened 2020-06-24
Added warning for developers using `pip install -e .`

dotsdl:
Adds a small warning for developers to the README; I got bit by this today.


jchodera:
Is there a way to remedy this issue by modifying `setup.py`?

dotsdl:
Let me see what I can do. I'll add changes to this PR if I can make this work smoothly with minimal changes.

dotsdl:
I don't have a good fix for this beyond:
1. Migrating the ff files into the package structure tree, where they will end up on a normal install.
2. Removing the `os.unlink`s in the `finally` block in `setup.py`; basically, not cleaning up.

I've looked around to see if using `pip install -e .` introduces some hooks we could grab onto for making the `os.unlink`s conditional. I would have to do a deeper dive to find such a thing than I've done so far.

I don't mind not being able to `pip install -e .` personally; it's usually my first reflex, but this isn't the only package I know of that has idiosyncrasies around that. The note I've added to the README satisfies me.

mattwthompson:
I've ~fixed this in #230/#227 by just including the files in the module structure directly. This might add some weight in packaging, but there are ways around that.

PR #145, opened 2021-01-21
Migrate to the C-F OpenFF Toolkit Package

SimonBoothroyd:
## Description

This PR handles the migrations needed to use the `openff-toolkit` package (formerly named `openforcefield`)  which is now available on conda-forge. 

The only real difference between the `openforcefield` package previously used here and the new `openff-toolkit` package is all instances of `import openforcefield.XXX` and `from openforcefield.XXX` are now replaced by `import openff.toolkit.XXX` and `from openff.toolkit.XXX`

## Status

- [x] Ready to go


SimonBoothroyd:
cc @jchodera

SimonBoothroyd:
CI won't run on this as secrets (the OE license) cannot be accessed from a fork. The `test_amber_import.py`, `test_template_generators.py`, `test_system_generator.py` pass locally (OSX).

jchodera:
Thanks! And wow, this namespace change is going to require a _ton_ of code changes everywhere.

PR #146, opened 2021-01-23
Replace encrypted license file

jchodera:
This updates the expired encrypted license file, which is the likely reason for why the tests were timing out.


jchodera:
Looks like `travis encrypt-file` is just dead.

PR #149, opened 2021-01-23
Update CI badge for GitHub Actions

jchodera:
This PR updates the CI badge in the README for GitHub Actions.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -80,9 +80,10 @@ jobs:
       shell: bash -l {0}
       run: |
 
-        # TODO: Find a way to avoid timing out on travis when running full charmm36.yaml conversion below
+        # TODO: Uncomment these tests when new ParmEd is released
+        # TODO: Find a way to avoid timing out when running full charmm36.yaml conversion below
         # python convert_charmm.py --verbose --in files/waters.yaml && python convert_charmm.py --verbose --in files/charmm36.yaml
-        python convert_charmm.py --verbose --in files/waters.yaml
+        # python convert_charmm.py --verbose --in files/waters.yaml
         # python convert_charmm.py --verbose
 
         # TODO: Uncomment this when tests are expected to work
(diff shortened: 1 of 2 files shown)


jchodera:
@SimonBoothroyd : I've added the `OE_LICENSE` secret and am re-running the tests.

Is [this warning](https://github.com/openmm/openmmforcefields/runs/1755136739?check_suite_focus=true#step:3:25) something to be worried about?
```
Warning: "channels" set on the "environment-file" do not match "channels" set on the action!
```

jaimergp:
Looks like a false warning, since they do match. `setup-miniconda` is a bit alarming with warnings but this is fine.

jchodera:
OK. Something is still failing due to a ParmEd issue. Looking into this now.

SimonBoothroyd:
> OK. Something is still failing due to a ParmEd issue. Looking into this now.

I had a quick look as well and essentially the last release of ParMed (3.2.0 which was released 2019) does not include the new `unique_atom_types` parameter to `OpenMMParameterSet.from_parameterset` which was added by Peter into master in April 2020.

Presumably Peter was working off of the master version of ParMed when putting together #117 and this wasn't caught earlier because the tests [timed out on travis](https://travis-ci.org/github/openmm/openmmforcefields/jobs/755377164) before it could reach the charm tests.

I guess the two options are to either drop the kwarg for now (which may lead to incorrect results with Drude?) or to try and push for a new ParMed + AmberTools release.

jaimergp:
Note that AmberTools [vendors a later version](https://github.com/conda-forge/ambertools-feedstock/blob/master/recipe/patches/parmed_version.py). Might not be late enough though.

SimonBoothroyd:
Ah that's good to know - it looks like it vendors February 2020 while this would need mid April 2020.

jaimergp:
I've been lobbying so parmed they _depend_ on ParmEd instead of vendoring random commits (or at least cutting a release for the vendored commit), but I ran out of energy to pursue that...

jchodera:
@SimonBoothroyd : It looks like the new ParmEd version is only required for CHARMM conversion, not the template or system generators, so I'll cut a new release with a warning now while we wait for ParmEd to cut and push a new release.

PR #156, opened 2021-02-27
Enable GLYCAM06j-1 forcefield conversion

zhang-ivy:
Convert `leaprc.GLYCAM_06j-1` to an opemm ffxml.

Requires https://github.com/ParmEd/ParmEd/pull/1149

~~TODO:~~
~~- [ ] The protein atoms in `GLYCAM_06j-1` are named `N`, `CA`, etc. Need to decide whether we should use `ff14SB` to match the existing protein atom names or enforce use of `protein.ff14SB` (which will require changing the protein atom names in `GLYCAM_06j-1` to `protein-N`, `protein-CA` etc).~~
~~- [x] Allow `override_level` parameter in yaml input to `convert_amber.py`~~
~~- [ ] Write test for validating energies (openmm vs tleap system) -- I've started this, but its not working. See below.~~

~~Currently, the primary showstopper for the test is that `ffxml/GLYCAM_06j-1.xml` does not contain residues that are actually part of glycans. The only residues that are present are the modified protein residues to which glycans can bond (e.g. NLN, which is the modified version of ASN).
For example, here are some of the residues that should be present: [‘UYB’, ‘4YB’, ‘VMB’, ‘2MA’, ‘0YB’, ‘0FA’, ‘0LB’]~~

~~This problem shows up as an error when I run `python convert_amber.py -i glycan.yaml` at
https://github.com/zhang-ivy/openmmforcefields/blob/convert-glycans/amber/convert_amber.py#L626~~

```
ValueError: No template found for residue 197 (UYB).  The set of atoms is similar to C5, but it is missing 3 atoms.
```

## Summary of the conversion pipeline:
1. In `openmmforcefields/amber/glycam/`, convert the glycam prep file (GLYCAM_06j-1.prep) into a .lib file (GLYCAM_06j-1.lib) 
- Make a tleap.in file to do this:

```python
# Read lines from prep file
with open("GLYCAM_06j-1.prep", "r") as f:
    lines_in = f.readlines()

# Make a list of residues
residues = []
# d_enantiomers = []
for line in lines_in:
    if "INT 0" in line:
        residue = line[:3]
        residues.append(residue)

# Remove CA2, as this isn't present as an atom type in ff14SB.xml or GLYCAM06j-1.xml or protein.ff14SB.xml or tip3p_standard.xml
residues.remove('CA2')

# Prep lines to write to tleap in file
lines = ["loadamberprep GLYCAM_06j-1.prep\n"]
for i in range(200, len(residues), 200): # Write 200 residues per line 
    lines.append("saveOff {" + " ".join(residues[i-200:i]) + "} GLYCAM_06j-1.lib\n")
lines.append("quit\n")

# Write lines
with open("make_glycam_lib.in", "w") as f:
    f.writelines(lines)
```
- Then run `tleap -s -f make_glycam_lib.in > make_glycam_lib.out`
- Output: GLYCAM_06j-1.lib

2. In `openmmforcefields/amber/glycam/`, create a modified glycam leaprc file (that replaces loading the .prep with loading the .lib)
    - Output: leaprc.GLYCAM_06j-1

3. In `openmmforcefields/amber/glycam/`, create a glycan.yaml and make sure to load the modified leaprc.

4. From `openmmforcefields/amber`, run `python convert_amber.py -i glycam/glycan.yaml --verbose`
    - Note you will see warnings like this that are safe to ignore. ParmEd thinks that there are a lot of identical residues because it doesn’t take into stereochemistry or anomeric forms
 
```
/home/zhangi/miniconda3/envs/glycan-ff-2/lib/python3.8/site-packages/ParmEd-3.4.0+21.gffe7aaf-py3.8-linux-x86_64.egg/parmed/openmm/parameters.py:666: UserWarning: Residue <ResidueTemplate YzB: 18 atoms; 18 bonds; head=C1; tail=O2> will be considered by OpenMM to be identical to <ResidueTemplate YZB: 18 atoms; 18 bonds; head=C1; tail=O2>.
warnings.warn(f'Residue {residue} will be considered by OpenMM to be identical to {residue_collision}.')
```

5. Use the converted glycam ffxml to parametrize a system: 
```python
from simtk import openmm
from simtk.openmm import app

# Load the ffxml
ffxml = ['../ffxml/protein.ff14SB.xml', "../ffxml/GLYCAM_06j-1_modified.xml"]

ff = app.ForceField(*ffxml)

# Load the test pdb
pdb = app.PDBFile("../test.pdb")

# Create system
system_omm = ff.createSystem(pdb.topology)
```

## Summary of changes to existing codebases:
**Parmed**
- Allow some 1-4 interactions to be unscaled: https://github.com/ParmEd/ParmEd/pull/1149
- Base unscaled pairs on all atoms in a torsion: https://github.com/ParmEd/ParmEd/pull/1182
- Perceive external bonds: https://github.com/ParmEd/ParmEd/pull/1184 (not merged yet)
**Openmmforcefields**
- Enable GLYCAM06j-1 forcefield conversion: https://github.com/openmm/openmmforcefields/pull/156
**OpenMM**
- Allow custom template matchers for residues: https://github.com/openmm/openmm/pull/3161
- Support initialization scripts in force fields: https://github.com/openmm/openmm/pull/3167


zhang-ivy:
@jchodera @mikemhenry @rafwiewiora : could you help me identify why there are no glycan residues in `ffxml/GLYCAM_06j-1.xml`?
@peastman and @jchodera : could you help me decide whether we should enforce protein atom names in `ffxml/GLYCAM_06j-1.xml` to contain the `protein-` prefix (to be used with `protein.ff14SB`)? Or should we leave `ffxml/GLYCAM_06j-1.xml` as is and use `ff14SB` as the protein ff?

mikemhenry:
Not sure if it helps or not but this site seems to have the parameter files:
http://glycam.org/docs/forcefield/all-parameters/
And the missing residues are in this file:
http://glycam.org/docs/forcefield/wp-content/uploads/sites/6/2014/03/GLYCAM_06j-1.prep

jchodera:
We managed to [get to the root of the issue](https://github.com/ParmEd/ParmEd/pull/1149#issuecomment-787216534), and I will try a full conversion today!

zhang-ivy:
I was able to address the "primary showstopper" (i.e. getting glycan residues into the ffxml, described in detail above) with the help of [John's comment](https://github.com/ParmEd/ParmEd/pull/1149#issuecomment-787216534). What I did:
1. Converted the glycam .prep file into a .lib file (using `tleap -s -f amber/glycam/make_glycam_lib_trunc.in > make_glycam_lib_trunc.out`): `glycam/GLYCAM_06j-1.lib`
2. Created a modified glycam leaprc file (that replaces loading the .prep with loading the .lib): `glycam/leaprc.GLYCAM_06j-1_trunc`
3. Updated `glycam/glycan.yaml` to load the modified leaprc.
4. From `amber/`, ran: `python convert_amber.py -i glycam/glycan.yaml --verbose`

This generated `ffxml/GLYCAM_06j-1_trunc.xml`. Note that this ffxml is truncated because it only includes the following glycan residues (which are present in RBD:ACE2): ['UYB', '4YB', 'VMB', '2MA', '0YB', '0fA', '0LB']. I initially tried to include all glycan residues in the ffxml, but came across many problems related to OpenMM perceiving different templates as the same, so I figured we could deal with those issues later. 

After generating the ffxml, I was getting errors like: 
`ValueError: No template found for residue 197 (UYB). The set of atoms matches UYB, but the bonds are different.` 
This was because the `ExternalBonds` in the ff templates were not correct. In `GLYCAM_06j-1_trunc.xml`, for the UYB residue, I was seeing: 
```
     <ExternalBond atomName="C1"/>
     <ExternalBond atomName="O4"/>
```
But there should be a third line for `atomName="O6".` Since Parmed was not writing out all of the atoms with external bonds correctly, I added a temporary fix to `parmed/openmm/parameters.py`. This fix involved adding a function called ` _get_atoms_with_external_bonds()` that gets atoms with external bonds.

Here is the function:
```
def _get_atoms_with_external_bonds(self, residue_name):
        # Determine the atoms with external bonds
        d = {'C': 4, 'N': 3, 'O': 2, 'H': 1} # Key: atom element name, Value: number of bonds it should have
        external_bonds = []
        for atom in self.residues[residue_name].atoms:
            bonds = 0

            # Add an extra bond if the atom is the O in a carbonyl
            if atom.name in ['OD1', 'O', 'O2N', 'OXT']:
                bonds += 1

            for bond in self.residues[residue_name].bonds:
                if bond.atom1 == atom or bond.atom2 == atom:
                    bonds += 1

                # Add extra bonds if the atom is the C in a carbonyl
                if (bond.atom1.name == 'OD1' and bond.atom2 == atom) or (bond.atom1 == atom and bond.atom2.name == 'OD1'):
                    bonds += 1
                elif (bond.atom1.name == 'O' and bond.atom2 == atom) or (bond.atom1 == atom and bond.atom2.name == 'O'):
                    bonds += 1
                elif (bond.atom1.name == 'O2N' and bond.atom2 == atom) or (bond.atom1 == atom and bond.atom2.name == 'O2N'):
                    bonds += 1
                elif (bond.atom1.name == 'OXT' and bond.atom2 == atom) or (bond.atom1 == atom and bond.atom2.name == 'OXT'):
                    bonds += 1
            if atom.element_name == 'Og': # If the atom is in a glycan, the element names are not set properly
                element_name = atom.name[0]
            else:
                element_name = atom.element_name
            if d[element_name] != bonds:
                external_bonds.append(atom)

        return external_bonds
```
And I added the following lines after [this line](https://github.com/ParmEd/ParmEd/blob/1be8ca0f2f5b232db79384cb02454a9c53f37749/parmed/openmm/parameters.py#L655):
```
external_bonds = self._get_atoms_with_external_bonds(name)
            for atom in external_bonds:
                if atom != residue.head and atom != residue.tail:
                    etree.SubElement(xml_residue, 'ExternalBond', atomName=atom.name)
``` 

After implementing this fix, I re-ran `python convert_amber.py -i glycam/glycan.yaml --verbose` and got an error in the [test](RBD_6m0j_refine_14_capped_N343glycosylated_cleaned_nowat.pdb) I wrote in `convert_amber.py`. The error occurs [here](https://github.com/zhang-ivy/openmmforcefields/blob/convert-glycans/amber/convert_amber.py#L631) and the message is:
```
parmed.exceptions.ParameterError: Cannot find necessary parameters
```
My temporary fix: At this [line](https://github.com/zhang-ivy/openmmforcefields/blob/convert-glycans/amber/convert_amber.py#L632), I replace `parm_omm` with `parm_amber`.

When I re-run with this fix, I can compare the omm vs amber energies:
```
amber energies [('HarmonicBondForce', 786.2137738648838), ('HarmonicAngleForce', 1573.0058453514364), ('PeriodicTorsionForce', 9090.943943094859), ('PeriodicTorsionForce', 13.594734381835927), ('NonbondedForce', -15185.572665049249), ('CMMotionRemover', 0.0)]

omm energies [('HarmonicBondForce', 785.9383122713264), ('HarmonicAngleForce', 1567.5857045366686), ('PeriodicTorsionForce', 9034.076934658806), ('NonbondedForce', -4528.287677760537), ('CMMotionRemover', 0.0)]
```
Notice there are two PeriodicTorsionForces in the amber energies..? And the NonbondedForce energies differ by a lot..

amber/convert_amber.py
```diff
@@ -366,7 +369,8 @@ def convert_yaml(yaml_name, ffxml_dir, ignore=ignore):
         source = provenance['Source'] = []
         for source_file in source_files:
             if MODE == 'LEAPRC':
-                _filename = os.path.join(AMBERHOME, 'dat/leap/cmd', source_file)
```
jchodera, line 369: We need to generalize this. Perhaps we can first try local path and then try $AMBERHOME if that fails?
```python
if os.path.exists(source_file):
   # use local
else:
   #use AMBERHOME

File: amber/convert_amber.py [hunk 2]
```diff
@@ -612,16 +622,17 @@ def assert_energies(prmtop, inpcrd, ffxml, system_name='unknown', tolerance=2.5e
 
     if openmm_topology is not None:
         system_omm = ff.createSystem(openmm_topology)
-        parm_omm = parmed.openmm.load_topology(openmm_topology, system_omm,
-                                               xyz=openmm_positions)
+        #parm_omm = parmed.openmm.load_topology(openmm_topology, system_omm,
```
jchodera, line 625: This is problematic---remove comments

amber/convert_amber.py
```diff
@@ -1247,6 +1258,68 @@ def validate_merged_lipids(ffxml_name, leaprc_name):
             os.unlink(f[1])
     if verbose: print('Lipids energy validation for %s done!' % ffxml_name)
 
+def validate_glyco_protein(ffxml_name, leaprc_name,
```
jchodera, line 1261: Use this test idea?

https://github.com/ParmEd/ParmEd/pull/1149#issuecomment-782546139

jchodera:
> When I re-run with this fix, I can compare the omm vs amber energies:

@zhang-ivy Can you paste the XML file you generated as a gist?

zhang-ivy:
@jchodera : Here are the system xmls ( `amber.xml` and `omm.xml`): https://gist.github.com/zhang-ivy/0d7d192e0f8296d609b8d4d94e4098ce

jchodera:
> @jchodera : Here are the system xmls ( amber.xml and omm.xml): https://gist.github.com/zhang-ivy/0d7d192e0f8296d609b8d4d94e4098ce

Oh, sorry, I was hoping to see the ffxml!

zhang-ivy:
@jchodera : The ffxml is included in this PR: https://github.com/openmm/openmmforcefields/blob/b73b6b0e420a737bb315da79b1ba4a66397caf33/amber/ffxml/GLYCAM_06j-1_trunc.xml

zhang-ivy:
@peastman helped me identify that the reason why there is an [energy discrepancy](https://github.com/ParmEd/ParmEd/pull/1149#issuecomment-810430224) between the amber vs openmm-generated test systems is because the prefixes are not correct in `GLYCAM_06j-1_trunc.xml`. 

I think I'll want to modify [this function](https://github.com/openmm/openmmforcefields/blob/master/amber/convert_amber.py#L548) such that it iterates through all atom types in GLYCAM_06j.dat and if the atom type is present in ff14SB, add the `protein-` prefix, otherwise add the `glycan-` prefix.

@peastman and @jchodera  : Please let me know what other considerations I need to take into account in order to write the appropriate changes necessary to address this prefix issue.

peastman:
Let me try to go over all the issues we need to consider.  Hopefully I'm not missing any!

Prefixes were a compromise to deal with differences in how OpenMM and Amber think about atom types.  OpenMM expects them to be globally unique.  If you load two files together, they can't each define a type with the same name.  But Amber sometimes has multiple files listing the same type.  The solution was to add prefixes like `protein-`, `DNA-`, and `RNA-`.  That way if you need to include both protein and DNA in a single system, you don't get conflicts.

Each atom type definition also defines a class, which is the same name but without the prefix:

https://github.com/openmm/openmmforcefields/blob/ca9319f15d0c5d8565853c0f525b54373cfffae1/amber/ffxml/protein.ff14SB.xml#L8

When defining parameters, they can be defined either by `type="protein-C"` or `class="C"`.  It appears that in all cases they do the former:

https://github.com/openmm/openmmforcefields/blob/ca9319f15d0c5d8565853c0f525b54373cfffae1/amber/ffxml/protein.ff14SB.xml#L2823

and

https://github.com/openmm/openmmforcefields/blob/ca9319f15d0c5d8565853c0f525b54373cfffae1/amber/ffxml/DNA.bsc1.xml#L1113

These are basically the same definition, but one using protein types and the other using DNA types.  This allows you to load either just the protein force field or just the DNA force field, and it has all parameters it needs.  Or you can load both of them together, and you don't have to worry about duplicate or conflicting definitions.  In principle we could separate all the common definitions into a different file that would specify them by class rather than type.  That would be a substantial change though.

All of this works fine for protein, DNA, and RNA.  Those are separate molecules so they can each be parametrized independently.  But glycans are bonded to other molecules.  A single bond/angle/torsion might include both protein and glycan atom types.  We need to figure out how to handle that.

Atom types appear in one other place.  https://github.com/ParmEd/ParmEd/pull/1149 added a list of atom types that should have 1-4 interactions scaled differently.  That list appears in an embedded script generated by ParmEd.  We could make the conversion script modify them the same way it modifies types in other places, but it might be cleaner to modify ParmEd so it can generate types with correct prefixes right from the start.

peastman:
While working on this, I ran into what I think is a bug in the glycam force field.  `GLYCAM_06j.dat` contains these lines.

```
NH-Cg-Cg-Sm   1    0.45          0.0            -3.         SCEE=1.0 SCNB=1.0 Thiol-linkages
              1    0.25          0.0             2.         SCEE=1.0 SCNB=1.0
```

However, the atom type NH is not defined.  This is the only place it is mentioned anywhere in the file.  It also is not a standard type used in the protein, DNA, or RNA force fields.

peastman:
Based on our meeting today, I created a script to post-process the XML file.  It makes the following assumptions about the input file:

1. No prefixes have been added.
2. All parameters are specified by type, not class.
3. The bad torsion noted above has been removed.  (Otherwise it throws an exception.)

It creates a modified XML file with the following changes.

1. All atom type definitions that are not GLYCAM-specific are removed.
2. All parameter definitions that do not involve at least one GLYCAM-specific type are removed.
3. For standard (not GLYCAM-specific) atom types, it adds the `protein-` prefix.

Here's a first draft of it.

```python
import xml.etree.ElementTree as etree
import re
import textwrap

# Define atom types

protein_types = set(["C", "CA", "CB", "CC", "CN", "CR", "CT", "CV", "CW", "C*", "CX", "H", "HC", "H1", "HA", "H4", "H5",
                     "HO", "HS", "HP", "N", "NA", "NB", "N2", "N3", "O", "O2", "OH", "S", "SH", "CO", "2C", "3C", "C8"])
solvent_types = set(["HW", "OW", "Li+", "Na+", "K+", "Rb+", "Cs+", "F-", "Cl-", "Br-", "I-"])
glycam_types = set()
replacements = {}
for type in protein_types:
    replacements[type] = 'protein-'+type

# Process <AtomTypes>

tree = etree.parse('glycam.xml')
root = tree.getroot()
types = root.find('AtomTypes')
for type in types.findall('Type'):
    name = type.get('name')
    if name in protein_types or name in solvent_types:
        types.remove(type)
    else:
        glycam_types.add(name)
        replacements[name] = 'glycam-'+name
        type.set('name', replacements[name])

# Process <Residues>

residues = root.find('Residues')
for residue in residues.findall('Residue'):
    for atom in residue.findall('Atom'):
        atom.set('type', replacements[atom.get('type')])

# Process <HarmonicBondForce>

force = root.find('HarmonicBondForce')
for bond in force.findall('Bond'):
    types = [bond.get('type1'), bond.get('type2')]
    if any(t in glycam_types for t in types):
        bond.set('type1', replacements[types[0]])
        bond.set('type2', replacements[types[1]])
    else:
        force.remove(bond)

# Process <HarmonicAngleForce>

force = root.find('HarmonicAngleForce')
for angle in force.findall('Angle'):
    types = [angle.get('type1'), angle.get('type2'), angle.get('type3')]
    if any(t in glycam_types for t in types):
        angle.set('type1', replacements[types[0]])
        angle.set('type2', replacements[types[1]])
        angle.set('type3', replacements[types[2]])
    else:
        force.remove(angle)

# Process <PeriodicTorsionForce>

force = root.find('PeriodicTorsionForce')
for tag in ['Proper', 'Improper']:
    for torsion in force.findall(tag):
        types = [torsion.get('type1'), torsion.get('type2'), torsion.get('type3'), torsion.get('type4')]
        if any(t in glycam_types for t in types):
            torsion.set('type1', replacements[types[0]])
            torsion.set('type2', replacements[types[1]])
            torsion.set('type3', replacements[types[2]])
            torsion.set('type4', replacements[types[3]])
        else:
            force.remove(torsion)

# Process <NonbondedForce>

force = root.find('NonbondedForce')
for atom in force.findall('Atom'):
    type = atom.get('type')
    if type in glycam_types:
        atom.set('type', replacements[type])
    else:
        force.remove(atom)

# Update the unscaled types in the script

script = root.find('Script')
text = script.text
types = ', '.join('"%s"' % replacements[s] for s in sorted(glycam_types))
types = '\n    '.join(textwrap.wrap(types))
pattern = re.compile('unscaled_types = \[(.*\n)*?.*?\]')
script.text = pattern.sub('unscaled_types = ['+types+']', text)

tree.write('modified.xml')
```

zhang-ivy:
Thanks @peastman ! Excited to try that out. 
As discussed in our meeting on Monday, my task was to adapt the `tleap.in` file used to convert the `GLYCAM_06j-1.prep` to a `GLYCAM_06j-1.lib` to contain only one enantiomer of each residue. I've done that, but when I run `python convert_amber.py -i glycam/glycan.yaml --verbose`, I get an error. 

Here is what I did:
1. Wrote a python script that creates a tleap.in file that takes in `GLYCAM_06j-1.prep` and generates a `.lib` file. Grabs all residue names from `GLYCAM_06j-1.prep` and keeps only one enantiomer for each residue. Output: `make_glycam_lib_one_enantiomer.in`
Script:
```python

# Read lines from prep file
with open("leaprc_GLYCAM_06j-1_2014-03-14/GLYCAM_06j-1.prep", "r") as f:
    lines_in = f.readlines()

# Grab residues, keeping only one enantiomer of each
residues = []
for line in lines_in:
    residue = line[:3]
    if "INT 0" in line and residue.upper() not in residues: # Keep only one enantiomer
        residues.append(residue)

# Make list of lines to write to the tleap.in file
lines = ["loadamberprep GLYCAM_06j-1.prep\n"]
for i in range(200, len(residues), 200):
    lines.append("saveOff {" + " ".join(residues[i-200:i]) + "} GLYCAM_06j-1.lib\n")
lines.append("quit\n")

# Write lines
with open("make_glycam_lib_one_enantiomer.in", "w") as f:
    f.writelines(lines)
```
2.  Converted the glycam .prep file into a .lib file (using tleap -s -f make_glycam_lib_one_enantiomer.in > make_glycam_lib_one_enantiomer.out). Output: GLYCAM_06j-1.lib
3. Created a modified glycam leaprc file (that replaces loading the .prep with loading the .lib): leaprc.GLYCAM_06j-1_trunc
4. Updated glycan.yaml to load the modified leaprc.
5. From `amber/`, ran: `python convert_amber.py -i glycam/glycan.yaml --verbose`

Got error:
```
Converting ['./glycam/leaprc.GLYCAM_06j-1_trunc'] to ffxml...
Preprocessing the leaprc for GLYCAM_06j-1_trunc...
Converting to ffxml ffxml/GLYCAM_06j-1_trunc.xml...
leaprc addAtomTypes {
        { "C"   "C" "sp2" }
        { "Cg"  "C" "sp3" }
        { "Cy"  "C" "sp3" }
        { "Ck"  "C" "sp2" }
        { "CT"  "C" "sp3" }
        { "Cj"  "C" "sp2" }
        { "Cp"  "C" "sp3" }
        { "H"   "H" "sp3" }
        { "H1"  "H" "sp3" }
        { "H2"  "H" "sp3" }
        { "Ha"  "H" "sp3" }
        { "Hp"  "H" "sp3" }
        { "Hc"  "H" "sp3" }
        { "Ho"  "H" "sp3" }
        { "HW"  "H" "sp3" }
	{ "Ng"  "N" "sp2" }
        { "NT"  "N" "sp3" }
        { "N3"  "N" "sp3" }
        { "Oh"  "O" "sp3" }
        { "Os"  "O" "sp3" }
        { "O"   "O" "sp2" }
        { "O2"  "O" "sp2" }
        { "OW"  "O" "sp3" }
        { "Oy"  "O" "sp3" }
        { "S"   "S" "sp3" }
        { "Sm"  "S" "sp3" }
        { "P"   "P" "sp3" }
}
glycam_06 = loadamberparams GLYCAM_06j.dat
loadOff glycam/GLYCAM_06j-1.lib
loadOff GLYCAM_amino_06j_12SB.lib
loadOff GLYCAM_aminoct_06j_12SB.lib
loadOff GLYCAM_aminont_06j_12SB.lib
addPdbResMap {
        { 0 "OLS" "NOLS" } { 1 "OLS" "COLS" }
        { 0 "OLT" "NOLT" } { 1 "OLT" "COLT" }
        { 0 "OLP" "NOLP" } { 1 "OLP" "COLP" }
        { 0 "HYP" "NHYP" } { 1 "HYP" "CHYP" }
        { 0 "NLN" "NNLN" } { 1 "NLN" "CNLN" }
}
HOH = TP3  WAT = TP3
ionsff = loadamberparams frcmod.ionsjc_tip3p
/home/zhangi/miniconda3/envs/glycan-ff/lib/python3.9/site-packages/ParmEd-3.4.0+19.g310f15e.dirty-py3.9-linux-x86_64.egg/parmed/openmm/parameters.py:361: ParameterWarning: Some residue templates are using unavailable AtomTypes
  warnings.warn('Some residue templates are using unavailable AtomTypes',
  File "/lila/home/zhangi/choderalab/openmmforcefields/amber/convert_amber.py", line 1420, in <module>
    main()
  File "/lila/home/zhangi/choderalab/openmmforcefields/amber/convert_amber.py", line 113, in main
    convert_yaml(args.input, ffxml_dir='ffxml/')
  File "/lila/home/zhangi/choderalab/openmmforcefields/amber/convert_amber.py", line 434, in convert_yaml
    ffxml_name = convert_leaprc(files, ffxml_dir=ffxml_dir, ignore=ignore,
  File "/lila/home/zhangi/choderalab/openmmforcefields/amber/convert_amber.py", line 222, in convert_leaprc
    params.write(ffxml_name, provenance=provenance, write_unused=write_unused, improper_dihedrals_ordering='amber')
  File "/home/zhangi/miniconda3/envs/glycan-ff/lib/python3.9/site-packages/ParmEd-3.4.0+19.g310f15e.dirty-py3.9-linux-x86_64.egg/parmed/openmm/parameters.py", line 377, in write
    self._write_omm_residues(root, skip_residues, skip_duplicates,
  File "/home/zhangi/miniconda3/envs/glycan-ff/lib/python3.9/site-packages/ParmEd-3.4.0+19.g310f15e.dirty-py3.9-linux-x86_64.egg/parmed/openmm/parameters.py", line 689, in _write_omm_residues
    external_bonds = self._get_atoms_with_external_bonds(name)
  File "/home/zhangi/miniconda3/envs/glycan-ff/lib/python3.9/site-packages/ParmEd-3.4.0+19.g310f15e.dirty-py3.9-linux-x86_64.egg/parmed/openmm/parameters.py", line 645, in _get_atoms_with_external_bonds
    if d[element_name] != bonds:
KeyError: 'S'
```

Is this potentially related to the atom prefix issue that Peter pointed out?

zhang-ivy:
After fixing the external bonds error above, I realized that the approach we outlined in our meeting (keep only one enantiomer for each glycan residue) will not work.

When I run `python convert_amber.py -i glycam/glycan.yaml --verbose` after fixing the above error, I am not able to parametrize my test system (glycosylated RBD) in tleap because the glycan residues needed (`NLN UYB 4YB VMB 2MA 0YB 0fA 0LB`) are not all present in the `GLYCAM_06j-1.lib` file. 

However, if I do not filter out any residues when converting the `.prep` to a `.lib`, I have no issues parametrizing my system in tleap, but get the following errror when parametrizing in OpenMM:
```
  File "/lila/home/zhangi/choderalab/openmmforcefields/amber/convert_amber.py", line 637, in assert_energies2
    system_omm = ff.createSystem(parm_amber.topology)
  File "/home/zhangi/miniconda3/envs/glycan-ff/lib/python3.9/site-packages/simtk/openmm/app/forcefield.py", line 1177, in createSystem
    templateForResidue = self._matchAllResiduesToTemplates(data, topology, residueTemplates, ignoreExternalBonds)
  File "/home/zhangi/miniconda3/envs/glycan-ff/lib/python3.9/site-packages/simtk/openmm/app/forcefield.py", line 1359, in _matchAllResiduesToTemplates
    [template, matches] = self._getResidueTemplateMatches(res, data.bondedToAtom, ignoreExternalBonds=ignoreExternalBonds, ignoreExtraParticles=ignoreExtraParticles)
  File "/home/zhangi/miniconda3/envs/glycan-ff/lib/python3.9/site-packages/simtk/openmm/app/forcefield.py", line 977, in _getResidueTemplateMatches
    raise Exception('Multiple non-identical matching templates found for residue %d (%s): %s.' % (res.index+1, res.name, ', '.join(match[0].name for match in allMatches)))
Exception: Multiple non-identical matching templates found for residue 197 (UYB): UVA, UVB, UWA, UWB, UYA, UYB.
```
It seems that the UYB template actually matches many other residues (that do not just differ by upper/lower case as I had initially thought). 
For now, I'll proceed with testing Peter's fix on the ffxml generated for the small subset of glycan residues ((`NLN UYB 4YB VMB 2MA 0YB 0fA 0LB`)  in my RBD system. But in order to make a ffxml including all glycan residues, we'll need to think of a more clever approach for how to circumvent the template matching issue. 
@peastman  and @jchodera : Let me know if you have any thoughts on how to resolve this. If not, maybe we can discuss with the amber dev folks.

zhang-ivy:
I've just modified the truncated ffxml (with only a small subset of glycan residues) using Peter's code and tested the ffxml on a test system (glycosylated RBD): 
```python
from simtk.openmm import app

# Load the ffxml
ffxml = [ffxml/protein.ff14SB.xml', "ffxml/GLYCAM_06j-1_trunc_modified.xml"]
if isinstance(ffxml, str):
    ff = app.ForceField(ffxml)
else:
    ff = app.ForceField(*ffxml)

# Load the test pdb
pdb = app.PDBFile("../test.pdb")

# Create system
system_omm = ff.createSystem(pdb.topology)

```
The last line yields this error:
```
----------
ValueErrorTraceback (most recent call last)
<ipython-input-11-f9fd22c0d06a> in <module>
----> 1 system_omm = ff.createSystem(pdb.topology)

~/miniconda3/envs/glycan-ff/lib/python3.9/site-packages/simtk/openmm/app/forcefield.py in createSystem(self, topology, nonbondedMethod, nonbondedCutoff, constraints, rigidWater, removeCMMotion, hydrogenMass, residueTemplates, ignoreExternalBonds, switchDistance, flexibleConstraints, drudeMass, **args)
   1326 
   1327         for force in self._forces:
-> 1328             force.createForce(sys, data, nonbondedMethod, nonbondedCutoff, args)
   1329         if removeCMMotion:
   1330             sys.addForce(mm.CMMotionRemover())

~/miniconda3/envs/glycan-ff/lib/python3.9/site-packages/simtk/openmm/app/forcefield.py in createForce(self, sys, data, nonbondedMethod, nonbondedCutoff, args)
   2433         force = mm.NonbondedForce()
   2434         for atom in data.atoms:
-> 2435             values = self.params.getAtomParameters(atom, data)
   2436             force.addParticle(values[0], values[1], values[2])
   2437         force.setNonbondedMethod(methodMap[nonbondedMethod])

~/miniconda3/envs/glycan-ff/lib/python3.9/site-packages/simtk/openmm/app/forcefield.py in getAtomParameters(self, atom, data)
    926                 return result
    927             else:
--> 928                 raise ValueError('%s: No parameters defined for atom type %s' % (self.forceName, t))
    929 
    930         def getExtraParameters(self, atom, data):

ValueError: NonbondedForce: No parameters defined for atom type glycam-Cg
```

jchodera:
>When I run python convert_amber.py -i glycam/glycan.yaml --verbose after fixing the above error, I am not able to parametrize my test system (glycosylated RBD) in tleap because the glycan residues needed (NLN UYB 4YB VMB 2MA 0YB 0fA 0LB) are not all present in the GLYCAM_06j-1.lib file.

It's important to remember that tleap matches by residue name, but OpenMM matches by template. It sounds like we'll need to generate a _complete_ translation of the `.prep` file to `.lib` to feed into tleap _and_ a pruned version of the `.lib` file to feed into ParmEd to generate the OpenMM ffxml. (Alternatively, we could modify ForceField to permit multiple matches, or add a mode to ParmEd to prune duplicate templates.)

jchodera:
> except for CA2, which must be filtered out because it has atom type C0 which is not in ff14SB.xml, GLYCAM06j-1.xml, proteiin.ff14SB.xml, or tip3p_standard.xml

Weird! This sounds like another bug. Let's make sure to add this and the other potential bug in the `.prep` file to the list of issues to raise with the glycam folks.

jchodera:
> It seems that the UYB template actually matches many other residues that do not just differ by upper/lower case as I had initially thought, so my code for keeping only one enantiomer above is not going to work.

OK, this suggests we will really need to add a mode to ParmEd to filter out identical residue templates on export to ffxml, or perhaps add this to @peastman's post-processing script as a short-term workaround.

zhang-ivy:
>Weird! This sounds like another bug. 

Actually, it seems like `C0` is present as a protein atom type in [peter's script](https://github.com/openmm/openmmforcefields/pull/156#issuecomment-865442426). Just not sure why `CA2` is present in GLYCAM_06j-1 and not in ff14SB, as CA2 seems to correspond to a Ca2+ ion, meaning it should have no parameters that are specific to the glycam forcefield. Will make a note to bring this up in the chat with the amber folks.

>and the other potential bug in the .prep file to the list of issues to raise with the glycam folks.
Which other potential bug are you referring to here?

zhang-ivy:
>OK, this suggests we will really need to add a mode to ParmEd to filter out identical residue templates on export to ffxml, or perhaps add this to @peastman's post-processing script as a short-term workaround.

Yes, I'm just not sure how we want to do this.
1) I'm not sure I fully understand the glycan naming system: http://glycam.org/docs/forcefield/glycam-naming-2/ 
It seems that there is a set of core monosaccharides (Table 1), where lowercase indicates L- and uppercase indicates D-. The third letter indicates the anomeric configuration (α or β). 
But it seems like the naming system also allows encoding of ring form (pyranosyl or furanosyl) and occupied linkage positions (2-, 2,3-, 2,4,6-, etc.). For these two properties, I'm pretty sure we want there to be different parameters for different ring forms and linkage positions? In which case we would need ParmEd's residue template matcher to be able to distinguish between residues with these diff properties so that it can properly match the right residue template to a given residue?

peastman:
> Just not sure why CA2 is present in GLYCAM_06j-1 and not in ff14SB, as CA2 seems to correspond to a Ca2+ ion

Ions get defined along with water, since their parameters can be different for different water models.  Add it to `solvent_types` in my script.  I only included the ions that were present in your converted XML file, but it sounds like the full force field includes more of them.

zhang-ivy:
Sorry for the confusion. I just realized that peter's script contains `CO`, not `C0`. I'm not sure where `C0` comes from. We can ask the amber developers about this.

@peastman  When you get the chance, could you take a look at this and see if you can spot what's going wrong here: https://github.com/openmm/openmmforcefields/pull/156#issuecomment-867730540 ?

peastman:
Your converted force field has no parameters for any force.

```xml
  <HarmonicBondForce>
    </HarmonicBondForce>
  <HarmonicAngleForce>
    </HarmonicAngleForce>
  <PeriodicTorsionForce ordering="amber">
    </PeriodicTorsionForce>
  <NonbondedForce coulomb14scale="0.8333333333333334" lj14scale="0.5">
    <UseAttributeFromResidue name="charge" />
    </NonbondedForce>
```

Make sure the input to the script satisfies the requirements listed in https://github.com/openmm/openmmforcefields/pull/156#issuecomment-865442426.  In particular, it must not have had any prefixes added to it, and all parameters must be specified by type rather than class.

peastman:
I just had an offline discussion with @jchodera about this.  We came up with a possible approach to address some of the problems.  Let me describe it and you can see if you think it would work.

It sounds like we're going to need to match templates based on residue names.  GLYCAM uses different parameters for versions of residues that have identical molecular graphs, so we can't rely on elements and bonds.  We discussed using a template generator to address this, but after we talked I realized that won't work.  A template generator modifies the force field, registering a new template which will match the (previously unmatched) residue.  But in this case, we need to select between two different templates that *both* match the residue.  A template generator won't solve that problem.

Instead we could create a new "template matcher" object that overrides the default matching behavior.  It would be a function `f(forcefield, residue)` that returns a template to use for a specific Residue object.  You would register it by calling `registerTemplateMatcher()` on the ForceField.  When building a System, it would be called for each residue to select a template to use.  If it returned None, the standard logic would be used.

For this application, the implementation would be trivial.  The matcher would have a dict mapping names to templates.  It would implement the matching as

```python
def __call__(self, forcefield, residue):
  if residue.name in self.templates:
    return self.templates[residue.name]
  return None
```

This would mostly be a very clean solution.  It would require one extra step from the user, though.  After creating your ForceField, you would have to call

```python
forcefield.registerTemplateMatcher(GlycamTemplateMatcher())
```

It would be even better if we could skip that step and have it registered automatically when you loaded the XML file.  We already allow force fields to include a `<Script>` tag containing Python code, but those get executed at the end, after the System is already built.  A possibility is to add another tag, `<InitializationScript>`, which would be executed right after the force field was created.  That tag could define and register the template matcher.

Any thoughts on this?

zhang-ivy:
@peastman : The new "template matcher" object sounds like it would work! Would you be able to take a stab at implementing this and then I can help test it?

peastman:
Yep, I'll get to work on it.  It shouldn't take long.

peastman:
Template matchers are implemented in https://github.com/openmm/openmm/pull/3161.

zhang-ivy:
@peastman Great! Just to clarify how I should test this:
1. Convert the glycam prep file to a lib file (without filtering out any residues, except CA2)
2. Run `python convert_amber.py -i glycam/glycan.yaml --verbose` to generate the glycam ffxml.
3. Run [Peter's script ](https://github.com/openmm/openmmforcefields/pull/156#issuecomment-865442426) to fix the prefixes in the ffxml.
4. Update my test to use a template matcher when I parametrize the system in openmm. This will look something like (copied from the test in [openmm/openmm#3161](https://github.com/openmm/openmm/pull/3161)):
```python
        ff = ForceField(StringIO(xml))

        # Load a water box.
        prmtop = AmberPrmtopFile('systems/water-box-216.prmtop')
        top = prmtop.topology

        # Register a template matcher that selects a particular one.
        def matcher(ff, res):
            return ff._templates['HOH2']
        ff.registerTemplateMatcher(matcher)

        # It should now succeed in building a System.
        system = ff.createSystem(top)

```
But for my test, I'll have to write my own function to replace `matcher` above:
```python
def GlycamTemplateMatcher(self, forcefield, residue):
  if residue.name in self.templates:
    return self.templates[residue.name]
  return None
```
where `self.templates` is available upon loading the glycam ffxml into openmm (i.e. I don't have to construct this dict myself). Does this all sound right to you? 

Also, one thing I see that is inconsistent across [openmm/openmm#3161](https://github.com/openmm/openmm/pull/3161) and [the code in your comment](https://github.com/openmm/openmmforcefields/pull/156#issuecomment-870799618) is that the code in your comment defines `matcher` to require 3 parameters, like this `def __call__(self, forcefield, residue)`.  However,  [here](https://github.com/openmm/openmm/pull/3161/files#diff-9902b73fb4f57fc94de4c8f245e03b574d2068a708b006c9b14e07626a91379dR436) and [here](https://github.com/openmm/openmm/pull/3161/files#diff-7e10bd00e3ea4dd2e148b51d75c4bedff6a777d116ef99cf7869f276e8cd5973R977) it expects 2 parameters `matcher(self, res)`.

I'm guessing the `matcher` doesn't need to take in both `self` and `forcefield` and that I should change my code above to be defined as `def GlycamTemplateMatcher(forcefield, residue)` (no changes required in [openmm/openmm#3161](https://github.com/openmm/openmm/pull/3161) )?

peastman:
The template matcher can be any callable object.  A function is callable, so you can define it as

```python
def match(ff, residue):
  # Do something to return a template
```

But other objects can also be callable by defining a `__call__()` method.  You will probably define yours something like this:

```python
class GlycamTemplateMatcher(object):
  def __init__(self):
    # Construct a set with the names of all GLYCAM residues 
    self.glycam_residues = ...
  def __call__(self, ff, residue):
    if residue.name in self.glycam_residues:
      return ff._templates[residue.name]
    return None
```

If a residue's name is in the set of GLYCAM residues, it will be looked up by name.  For all others, it returns None so the standard matching behavior will be used.  You do, in some way, need to come up with that set of residue names.  `ff._templates` includes every template that has been registered, including the ones for standard amino acids, water, etc.

zhang-ivy:
@peastman Thanks! I've made some progress on the steps outlined in my [last message](https://github.com/openmm/openmmforcefields/pull/156#issuecomment-870911679). I got your prefix code working and now I'm currently trying to test your template matcher implementation. What's the best way to install `openmm` from master?
I tried running `python setup.py install` in my clone of `openmm`, but I'm getting errors like:
```
REMOVING "/home/zhangi/miniconda3/envs/glycan-ff/lib/python3.9/site-packages/simtk/openmm"
REMOVING "/home/zhangi/miniconda3/envs/glycan-ff/lib/python3.9/site-packages/simtk/unit"
ERROR: Set OPENMM_INCLUDE_PATH to point to the include directory for OpenMM
Exiting
```
after which i ran `export OPENMM_INCLUDE_PATH=/home/zhangi/choderalab/openmm/openmmapi/include`
but now I am getting: 
```ERROR: Set OPENMM_LIB_PATH to point to the lib directory for OpenMM
Exiting```
And I'm not sure what all the right paths I should specify are..

**Comment by zhang-ivy** [comment]
@peastman : What's the easiest way to test your changes in [openmm/openmm#3161](https://github.com/openmm/openmm/pull/3161) ? I am trying to install openmm from master by running `python setup.py install` in `/home/zhangi/choderalab/openmm/wrappers/python` in my cloned repo, but I'm running into issues.

I've set:
```
export OPENMM_INCLUDE_PATH=/home/zhangi/miniconda3/envs/glycan-ff/include
export OPENMM_LIB_PATH=/home/zhangi/miniconda3/envs/glycan-ff/lib/
```
but I'm still seeing this error:
```
copying openmm/app/data/charmm36/water.xml -> build/lib.linux-x86_64-3.9/openmm/app/data/charmm36
running build_ext
building 'openmm._openmm' extension
creating build/temp.linux-x86_64-3.9
creating build/temp.linux-x86_64-3.9/src
creating build/temp.linux-x86_64-3.9/src/swig_doxygen
gcc -pthread -B /home/zhangi/miniconda3/envs/glycan-ff/compiler_compat -Wl,--sysroot=/ -Wno-unused-result -Wsign-compare -DNDEBUG -fwrapv -O2 -Wall -fPIC -O2 -isystem /home/zhangi/miniconda3/envs/glycan-ff/include -fPIC -O2 -isystem /home/zhangi/miniconda3/envs/glycan-ff/include -fPIC -DMAJOR_VERSION=@OPENMM_MAJOR_VERSION@ -DMINOR_VERSION=@OPENMM_MINOR_VERSION@ -I/home/zhangi/miniconda3/envs/glycan-ff/include -I/home/zhangi/miniconda3/envs/glycan-ff/lib/python3.9/site-packages/numpy/core/include -I/home/zhangi/miniconda3/envs/glycan-ff/include/python3.9 -c src/swig_doxygen/OpenMMSwig.cxx -o build/temp.linux-x86_64-3.9/src/swig_doxygen/OpenMMSwig.o -std=c++11
gcc: error: src/swig_doxygen/OpenMMSwig.cxx: No such file or directory
error: command '/usr/bin/gcc' failed with exit code 1
```

**Comment by peastman** [comment]
Instructions on compiling from source are at http://docs.openmm.org/latest/userguide/library.html#compiling-openmm-from-source-code.  @jchodera are we still doing nightly builds?  The last update to https://anaconda.org/omnia-dev/openmm was over two months ago.

**Comment by jchodera** [comment]
> The last update to https://anaconda.org/omnia-dev/openmm was over two months ago.

Looks like it broke. I believe the infrastructure is here:
https://github.com/omnia-md/conda-dev-recipes

While the azure build badge shows a failed status, nearly all the nightly CI runs completed successfully---they just skipped the build.

@peastman: If I had to guess, I think it might be because the nightly build recipe [hard-codes version 7.5.1](https://github.com/omnia-md/conda-dev-recipes/blob/master/recipes/openmm/meta.yaml#L17) and we're already past that version?

**Comment by jchodera** [comment]
> Instructions on compiling from source are at http://docs.openmm.org/latest/userguide/library.html#compiling-openmm-from-source-code.

I'll note that these instructions are hard to follow, underspecified, and out of date. Last time I had to set up a build environment from scratch following them, it took me at least half a day. Hopefully, there's a better way.

**Comment by peastman** [comment]
> I'll note that these instructions are hard to follow, underspecified, and out of date.

But we're working to improve them!

Assuming you have a CUDA toolkit and C++ compiler installed, you should be able to install all the other required tools with

```
conda install -c conda-forge cmake doxygen swig jinja2 numpy cython
```

From the root directory of the repository type

```
mkdir build
cd build
ccmake ..
```

Configure the CMake options as described in the manual, then compile with

```
make PythonInstall
```

**Comment by zhang-ivy** [comment]
@jchodera I ran the command you sent: `conda install -c conda-forge -c omnia-dev/label/cuda102 -c omnia openmm` but it doesn't seem like the version of openmm in my env has Peter's latest template matcher changes.
Screen Shot 2021-07-01 at 10 35 04 AM

When I manually check the zipped distribution `linux-64/openmm-7.6-py38_cuda102_1.tar.bz2` at https://anaconda.org/omnia-dev/openmm/files, Peter's changes are there. Not sure why this is happening

**Comment by zhang-ivy** [comment]
Nevermind, starting from a fresh env and using the command you sent `conda install -c conda-forge -c omnia-dev/label/cuda102 -c omnia openmm` works!

**Comment by jchodera** [comment]
You generally have to force reinstall, or force downgrade / clear cache / upgrade. We should add instructions for that.

I think this might require `conda install --force-reinstall` and possibly a `conda clean -tipsy` beforehand.

**Comment by peastman** [comment]
> Do you want to add "bump release version in omnia-dev recipe" to the OpenMM release checklist so we don't forget next time?

Added!

**Comment by zhang-ivy** [comment]
@peastman:  We're getting closer! I was able to test out your template matcher code -- it seems to be working just fine (no errors when building the system). 
When I parametrize the same system (glycosylated RBD) with amber vs openmm, there is still a discrepancy in the NonbondedForce (amber energy listed first, then omm energy):
```
HarmonicBondForce 315.19927383252946 315.1992738325294
HarmonicAngleForce 630.6303917213042 630.6279803494542
PeriodicTorsionForce 3650.0810277256983 3650.0809397059006
NonbondedForce -6088.017237000398 -6672.377713699921
```
I'm attaching here the glycam ffxml I used to build the omm system, as well as the omm and amber system xmls. I wonder if something is wrong with the ffxml code for scaling 1-4 interactions?
[debug_glycan_ff.zip](https://github.com/openmm/openmmforcefields/files/6750012/debug_glycan_ff.zip)

**Comment by peastman** [comment]
Excellent!  Let me investigate and see what I can find.

**Comment by peastman** [comment]
It's definitely coming from the exceptions.  If I zero them out with

```python
for i in range(nb.getNumExceptions()):
    p1, p2, charge, sigma, epsilon = nb.getExceptionParameters(i)
    nb.setExceptionParameters(i, p1, p2, 0, sigma, 0)
```

the energies agree perfectly.  Now let's see what's causing the difference...

peastman:
Oops!  Silly mistake.  Originally `unscaled_types` was a list of atom types.  Then we decided it needed to be a list of pairs of types, not single types.  But my script in https://github.com/openmm/openmmforcefields/pull/156#issuecomment-865442426 set it to a list of single types.  Strangely, when it looks for a tuple in a list of strings, it never finds it...

Let me fix the script.

peastman:
I fixed it, but once I did I realized we may need to rethink how we handle the scaling.  The list includes every pair of atom types that appears in a GLYCAM torsion.  Some of those atom types are GLYCAM types, and others are standard Amber types.  Most of them include at least one GLYCAM type, but not all.

```
    ('protein-H', 'protein-H1'), ('protein-H', 'glycam-H2'), ('protein-H', 'glycam-Hp'),
    ('protein-H', 'protein-O2'), ('protein-H', 'glycam-Os'), ('protein-H1', 'protein-H1'),
```

The exact same atom types should be scaled when they appear in a standard residue, but not when they appear in a GLYCAM residue.

Now I'm thinking that instead of basing it on atom types, perhaps we need to base it on residues.  A pair should be unscaled if at least one of the atoms is a GLYCAM residue.  Does that seem like it would work?

It might be hard to get ParmEd to generate the code for that, because it associates scale factors strictly with torsions.  We may have to just fix the script by hand. :(

peastman:
Another solution could be to base it on the full list of atoms in each torsion.  Back when I first started working this, I posted a [list of assumptions](https://github.com/ParmEd/ParmEd/issues/1136#issuecomment-780189609), one of which was that the scale factors would only depend on the types of the two interacting atoms, not the other two atoms connecting them.  I was told that was correct.  However, looking through the .dat file for GLYCAM, I see torsions listed in which the two atoms on the end are standard types, and you need to include the middle atoms to see whether the 1-4 interaction should be scaled.  For example,

```
O -C -Cg-H1   1    0.0           0.0             3          SCEE=1.0 SCNB=1.0 2-Hydroxy ethanamide
```

If the third atom had type CX instead of Cg, that would be a standard backbone torsion.

zhang-ivy:
Great catch. It sounds like the right choice is your second suggestion of basing it on the full list of atoms in each torsion. This would mean checking if each atom in the torsion contains a glycam atom and if so, leave it unscaled, otherwise scale it. Does this sound right / easy to implement?

peastman:
After thinking about this some more, basing it on the full list of atoms in the torsion seems like the "right" answer to me.  That's how the original Amber files specify it.  If we want to be sure we're translating them correctly, specifying it in the same way should be the most robust.

It also turns out that not all GLYCAM torsions are unscaled.  It actually includes a few torsions that do get scaled.  For example,

```
H -Ng-C -H1   1   -2.50          0.0             2.         SCEE=1.2 SCNB=2.0 Parm94 - Corrected to remove phase shift
```

peastman:
Looks like we both posted the same answer at the same time!  It will involve revisions to ParmEd, but I don't think they'll be too difficult.  I can hopefully get a PR in tomorrow.

peastman:
The ParmEd changes are in https://github.com/ParmEd/ParmEd/pull/1182.  The updated script for fixing prefixes is below.  With these changes, I get excellent agreement in energy.

```python
import xml.etree.ElementTree as etree
import re

# Define atom types

protein_types = set(["C", "CA", "CB", "CC", "CN", "CR", "CT", "CV", "CW", "C*", "CX", "H", "HC", "H1", "HA", "H4", "H5",
                     "HO", "HS", "HP", "N", "NA", "NB", "N2", "N3", "O", "O2", "OH", "S", "SH", "CO", "2C", "3C", "C8"])
solvent_types = set(["HW", "OW", "Li+", "Na+", "K+", "Rb+", "Cs+", "F-", "Cl-", "Br-", "I-"])
glycam_types = set()
replacements = {}
for type in protein_types:
    replacements[type] = 'protein-'+type

# Process <AtomTypes>

tree = etree.parse('GLYCAM_06j-1_trunc.xml')
root = tree.getroot()
types = root.find('AtomTypes')
for type in types.findall('Type'):
    name = type.get('name')
    if name in protein_types or name in solvent_types:
        types.remove(type)
    else:
        glycam_types.add(name)
        replacements[name] = 'glycam-'+name
        type.set('name', replacements[name])

# Process <Residues>

residues = root.find('Residues')
for residue in residues.findall('Residue'):
    for atom in residue.findall('Atom'):
        atom.set('type', replacements[atom.get('type')])

# Process <HarmonicBondForce>

force = root.find('HarmonicBondForce')
for bond in force.findall('Bond'):
    types = [bond.get('type1'), bond.get('type2')]
    if any(t in glycam_types for t in types):
        bond.set('type1', replacements[types[0]])
        bond.set('type2', replacements[types[1]])
    else:
        force.remove(bond)

# Process <HarmonicAngleForce>

force = root.find('HarmonicAngleForce')
for angle in force.findall('Angle'):
    types = [angle.get('type1'), angle.get('type2'), angle.get('type3')]
    if any(t in glycam_types for t in types):
        angle.set('type1', replacements[types[0]])
        angle.set('type2', replacements[types[1]])
        angle.set('type3', replacements[types[2]])
    else:
        force.remove(angle)

# Process <PeriodicTorsionForce>

force = root.find('PeriodicTorsionForce')
for tag in ['Proper', 'Improper']:
    for torsion in force.findall(tag):
        types = [torsion.get('type1'), torsion.get('type2'), torsion.get('type3'), torsion.get('type4')]
        if any(t in glycam_types for t in types):
            torsion.set('type1', replacements[types[0]])
            torsion.set('type2', replacements[types[1]])
            torsion.set('type3', replacements[types[2]])
            torsion.set('type4', replacements[types[3]])
        else:
            force.remove(torsion)

# Process <NonbondedForce>

force = root.find('NonbondedForce')
for atom in force.findall('Atom'):
    type = atom.get('type')
    if type in glycam_types:
        atom.set('type', replacements[type])
    else:
        force.remove(atom)

# Update the unscaled types in the script

script = root.find('Script')
text = script.text
pattern = re.compile('unscaled_types = set\((\[(.*\n)*?.*?\])\)')
match = pattern.search(text)
types = eval(match.group(1))
types = [(replacements[t[0]], replacements[t[1]], replacements[t[2]], replacements[t[3]]) for t in types]
types = ',\n    '.join(str(t) for t in types)
script.text = pattern.sub('unscaled_types = set([%s])' % types, text)

tree.write('modified.xml')
```

zhang-ivy:
@peastman : I've just updated the testing pipeline with your changes to ParmEd and your prefix fixing code and I am also seeing excellent agreement (for RBD with glycans):
```
HarmonicBondForce 315.19927383252946 315.1992738325294
HarmonicAngleForce 630.6303917213042 630.6279803494542
PeriodicTorsionForce 3650.0810277256983 3650.0809397059006
NonbondedForce -6088.017237000398 -6088.017217994459
```
Hooray!! Thanks so much for all your help! 🎉 

The next steps for me are to: 
1) Test the converted ffxml to make sure we see good agreement in energies for a couple other systems.
2) Clean up the code in `convert_amber.py` and push the changes to this PR
3) Write up notes outlining the steps for converting the glycam ff to a ffxml and post them in this PR.

jchodera:
Woohoo!

peastman:
Sounds good!

zhang-ivy:
I've finished cleaning up this PR!

I've updated the [first comment in this PR](https://github.com/openmm/openmmforcefields/pull/156#issue-581255909) with a summary of the conversion pipeline and changes required by other repos. Note that I'm still waiting on https://github.com/ParmEd/ParmEd/pull/1184 to be merged.

I've also spent the afternoon modifying the `validate_glyco_protein` test in `convert_amber.py` to use the glycoprotein files that John suggested [here](https://github.com/ParmEd/ParmEd/pull/1149#issuecomment-782546139). In to get the tests working, I modified the files John initially sent and included them in this PR 
 (in `amber/files/glycam/Glycoprotein_shortened*`). Here are the changes I made:
- The .parm7 and .rst7 files were generated by a .leapin file that loads oldff/leaprc.ff10 and oldff/leaprc.GLYCAM_06j_10. In order to generate .parm7 and .rst7 files using updated forcefields that will be used in practice, I updated the .leapin file to load leaprc.ff14SB and leaprc.GLYCAM_06j-1 instead. 
- The .parm7 and .rst7 files include 7 molecules, but I removed the first 2 molecules (by editing the .leapin file) because they involve zwitterionic molecules, which I wasn't sure how to parametrize in OpenMM).

The test now outputs:
```
Validating the conversion...
Glycosylated protein energy validation for ffxml/GLYCAM_06j-1.xml
HarmonicBondForce 166.05257416728875 166.0525741672887
HarmonicAngleForce 443.64924500308035 443.6478171109116
PeriodicTorsionForce 548.348291428228 548.3482921829175
NonbondedForce 169459044.11620137 169459044.18473148
CMMotionRemover 0.0 0.0
```
I should note that I've set the energy tolerance to 1e-1 so that it passes. 

@jchodera @peastman : Could you please review and merge when you get a chance?

peastman:
Excellent, congratulations on getting this done!

zhang-ivy:
@peastman Thanks so much for your help!!

jchodera:
@peastman : We should cut a new release with the glycam functionality once you've had a chance to review and merge this.

peastman:
It looks good to me.  If all the tests pass, it should be good to merge.  (I notice that CI is currently failing due to lack of an OpenEye license key, but that's not related to this.)

jchodera:
>  (I notice that CI is currently failing due to lack of an OpenEye license key, but that's not related to this.)

Will fix!

jchodera:
@peastman : I've updated the OpenEye toolkit license and have triggered all the tests to run again.

zhang-ivy:
@jchodera : It looks like there's still a problem with the open eye license

jchodera:
The issue here is that secrets are not disclosed on PRs opened from forks.

I'll merge this and we can debug from here.

jchodera:
@zhang-ivy I've added you to this repo so you can use branches instead of pushing from forks. That should get around this issue next time.

PR #167, opened 2021-07-21
fix dependabot so it can see our secrets

mikemhenry:
Dependabot changed the way it accesses repo secrets to prevent a malicious package update from exfiltrating secrets. 

 https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-852541544


mikemhenry:
Not sure who is the right person to review this, but this should make it so we can test Dependabot version bumps

PR #180, opened 2021-10-28
Fix glycam ffxml and conversion script

zhang-ivy:
This PR makes the following changes:
- Fix problems with the external bonds for the following residues: terminal residues (e.g. NNLN, CNLN), OLP, and  OME.
    - Terminal residue issue pointed out [here](https://github.com/openmm/openmm/issues/3191#issuecomment-953239669) 
- Update the initialization script according to OpenMM API updates introduced by https://github.com/openmm/openmm/pull/3303

Requires: https://github.com/ParmEd/ParmEd/pull/1206

A couple remaining issues:
1. @peastman pointed out that the HYP residues does not use PDB v3 atom names [here](https://github.com/openmm/openmm/issues/3191#issuecomment-953335317). The hydrogens.xml I generated relies on GLYCAM_06j-1.xml to determine the hydrogens for each residue. I'm not sure why the HYP hydrogen names do not match the ones in CHYP/NHYP. Could you share your script for converting the HYP hydrogen names to PDB v3 names? Or maybe you could just make the changes directly to the updated GLYCAM_06j-1.xml included in this PR and commit them?
2. @peastman : I've fixed the bond perception issues for residues that I saw problems for and then chose a couple more random residues and checked thata those looked right. I'm not sure if there are more bond perception problems though, would you be able to randomly choose some residues and see if the external bonds look right?

### DIFF
--- a/amber/convert_amber.py
+++ b/amber/convert_amber.py
@@ -1493,19 +1493,37 @@ def modify_glycan_ffxml(input_ffxml_path):
 
     # Add initialization script for setting up GlycamTemplateMatcher
     initialization_script = etree.SubElement(root, 'InitializationScript')
-    initialization_script.text = """
+    initialization_script.text = """ 
+from openmm.app.internal import compiled
+
 class GlycamTemplateMatcher(object):
+
   def __init__(self, glycam_residues):
     self.glycam_residues = glycam_residues
-  def __call__(self, ff, residue):
+
+  def __call__(self, ff, residue, bondedToAtom, ignoreExternalBonds, ignoreExtraParticles):
     if residue.name in self.glycam_residues:
-      return ff._templates[residue.name]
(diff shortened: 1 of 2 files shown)


jchodera:
@peastman: Can you help us get this merged in since the beta is now out?

peastman:
What part do you need my help with?

zhang-ivy:
This should be good to merge!

mikemhenry:
@jchodera should we cut a patch release (eg 0.10.1) so this fix is on conda-forge?

jchodera:
> @jchodera should we cut a patch release (eg 0.10.1) so this fix is on conda-forge?

Apologies for missing this earlier! We've now got 0.11.0 out:
https://github.com/openmm/openmmforcefields/releases/tag/0.11.0

PR #191, opened 2022-03-28
Update for upcoming OpenFF Toolkit API breaks

mattwthompson:
Hi OpenMM developers,

I'm seeing some test failures in our example that uses `openmmforcefields` with our big biopolymer feature branch, probably due to `Atom.element` being removed in our upcoming API break. This change should fix that and be backwards-compatible.


devtools/conda-envs/test_env.yaml
```diff
@@ -27,9 +27,11 @@ dependencies:
     # JSON caching of residue templates
   - tinydb
 
-    # openff toolkit and force fields
-  - openff-toolkit >=0.7.1
+    # OpenFF toolkit and force fields
+  - openff-toolkit >=0.10.3|<=0.10.0
   - openff-forcefields >=1.2.0
+  - pip:
+    - git+https://github.com/openforcefield/openff-toolkit.git@topology-biopolymer-refactor
```
mattwthompson, line 34: If you don't want to shoulder the complexity of testing against an unreleased version, you can grab it with a similar command locally (`pip install git+https://github.com/openforcefield/openff-toolkit.git@topology-biopolymer-refactor`) and have a look around for yourself.

mattwthompson: And for the record: I'm not suggesting this change be merged in; I want to show in the context of this PR that it should pass with the new API. I will revert this before merging if approved.

openmmforcefields/generators/template_generators.py
```diff
@@ -236,9 +236,13 @@ def _generate_unique_atom_names(self, molecule):
...
         """
         from collections import defaultdict
+
+        # OpenFF Toolkit v0.11.0 removed Atom.element and replced it with Atom.symbol, etc.
+        uses_old_api = hasattr(molecule.atoms[0], "element")
+
         element_counts = defaultdict(int)
         for atom in molecule.atoms:
-            symbol = atom.element.symbol
+            symbol = atom.element.symbol if uses_old_api else atom.symbol
```
jchodera, line 245: Oh, that's an unexpected change.

@mattwthompson, can you point to some information about the motivation for this? Is the idea that `Atom` objects don't have to be physical elements anymore?

devtools/conda-envs/test_env.yaml
jchodera: Thanks for this! We should test against both the current and preview versions, ideally. But to do that, we can't just eliminate testing the current version---I'll have to rework the test matrix.

jchodera:
Thanks for this, @mattwthompson! 

I'll need to finish up https://github.com/openmm/openmmforcefields/pull/182 first, but then will tackle integrating this PR.

openmmforcefields/generators/template_generators.py
jchodera: And if `symbol` supports non-physical elements or made-up symbols now, is there an API for telling whether the element is a real element or not?

richardjgowers:
It would be great if this could get merged, this is breaking stuff for openfe currently

openmmforcefields/generators/template_generators.py
mikemhenry: This is the PR where the change happened: https://github.com/openforcefield/openff-toolkit/pull/1182
It comes from switching to a different package.

mikemhenry: @jchodera this is the only modification needed to get openmmforcefields working with the changes in the openff-toolkit/openff-element and given they are done in a backwards compatible way, I am happy with them. If you are happy with it, I can make a new PR add this changes + tests the latest version of the openff-toolkit.

jchodera (approved):
Thanks!

mikemhenry (approved):

mattwthompson:
Thanks!

mattwthompson:
Unfortunately our [tests are still](https://github.com/openforcefield/openff-toolkit/runs/7742439141?check_suite_focus=true) failing; I think L966 in this loop needs a similar change:

https://github.com/openmm/openmmforcefields/blob/a59b1445d8fe41c873c052590e330e615931e782/openmmforcefields/generators/template_generators.py#L960-L967

if I had to guess, I didn't notice this previous because I was grepping around for `atom.element.symbol` or something similar - and at first glance the traceback looks like `particle` is an OpenMM object of sorts, not an OpenFF `Atom` object.

PR #202, opened 2022-05-16
Fix CI

mikemhenry:
Looks like CI is failing with `The process '/usr/share/miniconda/condabin/conda' failed with exit code null` for the last 20 days, seeing if mamba will give us a better error


mikemhenry:
Going to merge this is in, looks like some espaloma failures, will need to debug if the issue is with the newer pytorch

PR #213, opened 2022-07-12
Fix #211 : Use basename instead of full path for cache database table names

### LINKED ISSUE #211: SMIRNOFFTemplateGenerator cache database table name should be file name not file path
If you pass a full path to to your offxml file to SMIRNOFFTemplateGenerator, this full path is used as the database table name in the cache. This is problematic for us as in our workflow this offxml file is a temporary file (we are running the calculations remotely in a temporary directory and we send the file data and write the file to the temporary directory ourselves), so the cache does not work properly. It would make more sense if the database table name was the file name, e.g.: 
`self._database_table_name = os.path.basename(forcefield)`

jchodera:
Fixes #211


jchodera:
There's a weird `osx` segfault we need to sort out, but will have to do this in a separate PR.

PR #217, opened 2022-07-15
Fix #216: Remove % sign from parmchk2 command

### LINKED ISSUE #216: Errant percentage sign in string stops parmchk2 from running successfully on some Windows machines
In the command string for parmchk2,
`cmd = f"parmchk2 -i out.mol2 -f mol2 -p {self.gaff_dat_filename} -o out.frcmod -s %{self._gaff_major_version}"`
 there is an extra percentage sign before the gaff major version that I think should not be there. On Linux/Mac, this seems to be harmless. Not entirely sure how this happens, but on Windows this sometimes seems to get filled with random text that gets inserted into the middle of the parmchk2 command that then does not run. Having tested on several different Windows machines, I only found this to happen on a single machine.

rosemary-cresset:
Unnecessary % sign in the parmchk2 command string sometimes gets filled
with junk characters on Windows and stops the command from running
successfully. This behaviour only occurs on some Windows systems and
does not occur on Linux or Mac.

### DIFF
--- a/openmmforcefields/generators/template_generators.py
+++ b/openmmforcefields/generators/template_generators.py
@@ -766,7 +766,7 @@ class GAFFTemplateGenerator(SmallMoleculeTemplateGenerator):
             _logger.debug(output)
 
             # Run parmchk.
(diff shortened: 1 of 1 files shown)


jchodera (approved):
Thanks for catching this!

PR #218, opened 2022-07-15
Fix #214: Leave temp dir before raising exceptions

### LINKED ISSUE #214: Recursive exception makes it very difficult to read acdoctor error message on Windows
Not sure if you officially support Windows, so feel free to close this if you don't. On Windows, if acdoctor finds a problem with the ligand, it is very difficult to read because the error message is swamped in hundreds of other error messages along these lines: 
During handling of the above exception, another exception occurred:\r\n\r\nTraceback (most recent call last):\r\n  File \"C:\\Users\\exper\\AppData\\Local\\Programs\\Cresset-BMD\\Flare\\python39.zip\\shutil.py\", line 629, in _rmtree_unsafe\r\nPermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\\\Users\\\\exper\\\\AppData\\\\Local\\\\Temp\\\\tmpdbm8egcl'\r\n\r\nDuring handling of the above exception, another exception occurred:\r\n\r\nTraceback (most recent call last):\r\n  File \"C:\\Users\\exper\\AppData\\Local\\Programs\\Cresset-BMD\\Flare\\python39.zip\\tempfile.py\", line 805, in onerror\r\nPermissionError: [WinError 5] Access is denied: 'C:\\\\Users\\\\exper\\\\AppData\\\\Local\\\\Temp\\\\tmpdbm8egcl'\r\n\r\n

This happens because the process changes to a temporary directory that it then tries to delete (after the exception is raised) without changing out of it again. This does not occur on Linux or Mac. To fix this, before both instances of `raise Exception(msg)` in template_generators.py, there should be a line `os.chdir(cwd)` to ensure that the temporary directories can be deleted.

rosemary-cresset:
Change out of temporary directory before raising an exception that
causes the directory to be deleted as the context manager exits. If not
done, a PermissionError is raised repeatedly on Windows and the
message from the original exception is very difficult to read. This
behaviour was observed for the parmchk2 command, but the fix has also
been applied to the antechamber command.

This fix could also have been implemented using try/finally - I can redo it with that if you prefer.

### DIFF
--- a/openmmforcefields/generators/template_generators.py
+++ b/openmmforcefields/generators/template_generators.py
@@ -762,6 +762,7 @@ class GAFFTemplateGenerator(SmallMoleculeTemplateGenerator):
                 msg += read_file_contents(local_input_filename)
                 msg += 8 * "----------" + '\n'
                 # TODO: Run antechamber again with acdoctor mode on (-dr yes) to get more debug info, if supported
+                os.chdir(cwd)
                 raise Exception(msg)
             _logger.debug(output)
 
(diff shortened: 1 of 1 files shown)


jchodera (approved):
This is great---thanks!

PR #219, opened 2022-07-15
Fix #215: Copy gaff dat file to temp dir

### LINKED ISSUE #215: parmchk2 fails to run if openmmforcefields is installed somewhere with spaces in the path
If the GAFF dat file is located somewhere with spaces in the path, as is common on Windows, the parmchk2 command from template_generators.py does not work. Adding quotes around the file path didn't help, but I found it could be fixed by copying the GAFF dat file to the temporary directory and passing that to parmchk2 instead.

rosemary-cresset:
Copy the gaff dat file to the temporary directory before running
parmchk2 to avoid problems with spaces in the path.

### DIFF
--- a/openmmforcefields/generators/template_generators.py
+++ b/openmmforcefields/generators/template_generators.py
@@ -766,7 +766,8 @@ class GAFFTemplateGenerator(SmallMoleculeTemplateGenerator):
             _logger.debug(output)
 
             # Run parmchk.
-            cmd = f"parmchk2 -i out.mol2 -f mol2 -p {self.gaff_dat_filename} -o out.frcmod -s %{self._gaff_major_version}"
+            shutil.copy(self.gaff_dat_filename, os.getcwd())
+            cmd = f"parmchk2 -i out.mol2 -f mol2 -p {os.path.basename(self.gaff_dat_filename)} -o out.frcmod -s %{self._gaff_major_version}"
             _logger.debug(cmd)
             output = subprocess.getoutput(cmd)
             if not os.path.exists('out.frcmod'):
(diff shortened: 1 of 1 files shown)


openmmforcefields/generators/template_generators.py
```diff
@@ -766,7 +766,8 @@ def read_file_contents(filename):
             _logger.debug(output)
 
             # Run parmchk.
-            cmd = f"parmchk2 -i out.mol2 -f mol2 -p {self.gaff_dat_filename} -o out.frcmod -s %{self._gaff_major_version}"
+            shutil.copy(self.gaff_dat_filename, os.getcwd())
```
jchodera, line 769: It occurs to me that there may also be issues if the *filename* has spaces in it a well. Would it be possible to just use a temporary name `gaff.dat` instead of `os.path.basename(self.gaff_dat_filename)`, so that these would be changed to
```python
            shutil.copy(self.gaff_dat_filename, 'gaff.dat')
            cmd = f"parmchk2 -i out.mol2 -f mol2 -p gaff.dat -o out.frcmod -s %{self._gaff_major_version}"
```

jchodera (changes requested):
@rosemary-cresset : Thanks so much for this, and apologies for taking so long to get to reviewing this! We'll be working on getting a new release out the door quickly!

openmmforcefields/generators/template_generators.py
rosemary-cresset: Good point. :-) I've made the change as you suggested.

jchodera:
@mikemhenry : This should be merged if the ubuntu tests pass.

jchodera:
@mikemhenry : Tests finally passed, so I've merged this!

PR #226, opened 2022-08-12
add back in openeye toolkit

mikemhenry:


mikemhenry:
This is segfaulting on osx b/c of espaloma, but this isn't a new problem. Since it speeds up tests, I'm going to merge this in

PR #227, opened 2022-08-23
Fix toolkit tests

mikemhenry:


mattwthompson:
The Python 3.8 failure is cryptic but the 3.10 failures look "correct" to me, and they fail in 10 minutes compared to  30-40, which implies to me OpenEye Toolkits are actually being used

mikemhenry:
Supersedes #225

mattwthompson:
I'm looking to find a more stable and portable way to ship the XML files. I can't for the life of me figure out why the symlinks are going two directories up? It looks like these paths are in the root of the repo.

https://github.com/openmm/openmmforcefields/blob/d4e6a07d89ddaef53371f605146ab6e6714d76db/setup.py#L24-L26

mikemhenry:
I do think the root of a lot of issues is not packing data files correctly, we should probably switch to a modern approach 
https://setuptools.pypa.io/en/latest/userguide/datafiles.html

mattwthompson:
The thing is, I'm pretty sure the package data is configured properly, or at least uses an approach that's on the better end of current best practices:

https://github.com/openmm/openmmforcefields/blob/d4e6a07d89ddaef53371f605146ab6e6714d76db/setup.py#L59-L73

Exactly why symlinks should be used is not clear to me; they're not portable and it saves no space on disk. I've attempted to remove them altogether with #229

mattwthompson:
@mikemhenry could you re-enable Codecov for this repo? https://app.codecov.io/gh/openmm/openmmforcefields

mikemhenry:
@mattwthompson I will see what I can do about codecov, I might need to get my permissions for this repo elevated

mikemhenry:
I requested org access, not sure if @peastman or @jchodera will get the email.

peastman:
I got the request.  What exactly am I approving?  The description says, "Granting access will give this application the ability to request access to private data in the openmm organization."  That's both vague and dangerous sounding.

mattwthompson:
I think the ["GitHub App"](https://github.com/marketplace/codecov) is what needs to be installed (or maybe authorized, or maybe something else ...). Lee-Ping actually ran into the same confusing interface last week while granting permissions to a repo (under his personal account) - and of course [you're not alone in reading the fine print](https://community.codecov.com/t/full-control-of-private-repositories/3442). He was confused for the same reasons. He was able to find a way to grant it only access to public repos - I don't know if either of you see this drop-down:

image

He also wrote:


I got some strange behavior but I finally resolved the issue.
1. I tried logging into Codecov through a private window and clicking "Public repos only" from the dropdown menu as shown.  I was able to log in by entering my GitHub credentials without the additional permission request.
2. I tried logging out of GitHub in my normal window and logging into CodeCov the same way as in step (1), but I got the additional permission request.
3. I cleared all my browser data and I was able to log in using a normal window, same as the private window.
The summary is there must have been something funny going on in my browser cache .. and unless you want to give Codecov full read/write access to all repos, it's better to click "Public repos only" from the dropdown menu.

jchodera:
@peastman: Continuous integration test coverage evaluation with `codecov` is a [best practice recommended by the Molecular Sciences Software Institute](https://education.molssi.org/python-package-best-practices/08-CI/index.html):
image

peastman:
> Continuous integration test coverage evaluation with codecov is a [best practice recommended by the Molecular Sciences Software Institute](https://education.molssi.org/python-package-best-practices/08-CI/index.html):

But granting untrusted applications unrestricted access to your organization's private data is very much not a best practice!

I'll see if I can figure out the method Lee-Ping described.

peastman:
Is one of you able to sign up for the Codecov account?  They have really daunting terms of service and privacy policy.  Those aren't things I can agree to without reading them first, and that's likely to take quite a while.  And just a quick skim brings up some red flags, e.g. there's a whole section on indemnification.

mattwthompson:
Let me kick the tires on an alternative I've been meaning to try

jchodera:
I have logged into codecov.com via "Public repos only" and granted access to the openmm org.

jchodera:
I've added `CODECOV_TOKEN` as a repository secret environment variable.

Do I also need to connect codecov integration in a separate step? The [link](https://app.codecov.io/gh/openmm/openmmforcefields) you provided notes "There was a problem getting repo contents from your provider", though it may still be syncing repos.

mattwthompson:
If I can grab the token from an org secret, I don't think you need any more setup. I'll give this a few more kicks tomorrow to be sure.

mattwthompson:
After excessive and inefficient iterations on my part, I think this is ready for review. It includes at least the following changes:

* Fixing a missed case from #191
* In general, gets tests passing with both the old and new APIs of the OpenFF Toolkit
* Updates some tests and iterator names to avoid referring to particles in OpenFF `Molecule` objects, which do not have the concept of particles
* Switches CI to use `micromamba`
* Fixing how package data is handled #230
* Adds testing on Python 3.10 (macOS only) #234
* Brings code coverage reports online
  * The reports' diffs might not be useful until the [default branch is updated in CodeCov's settings](https://github.com/openmm/openmmforcefields/pull/236#issuecomment-1247502457)
* Updates the `.gitignore` file to make my screen less red
* Updates some Python 2 cruft and sorts imports
* Removes a stray `simtk` import
* Adds `openmmforcefields/tests/__init__.py` in order for tests to be recognized as a module


There's more I'd like to do, but this is already borderline unreviewable due to my lack of restraint. I can't do much more than point to that the automation has been improved and tests are passing.

mattwthompson:
I can't request a review via the UI since you opened it, so a ping will have to do @mikemhenry

mattwthompson (approved):
I approve (I think?)

mikemhenry:
I'll review this PR after lunch to just double check it, but it should be good to merge in, like we don't have others we want to merge in first, right? @mattwthompson

mattwthompson:
Nope, the other PRs have either been merged into this one or were experiments that didn't work out. It's just this one (for now)!

.github/workflows/ci.yaml
```diff
@@ -52,25 +55,30 @@ jobs:
     - name: "Install latest openff-toolkit"
       if: ${{ matrix.latest-openff-toolkit == true }}
       run: |
-        # Until #225 is complete, we should not assume compatibility with new API
-        micromamba install "openff-toolkit =0.10.6" -c conda-forge -y
+        micromamba update -y -c conda-forge "openff-toolkit >=0.11.0"
 
     - name: Install Package
       run: |
         # Can't use 'develop' mode because we remove symlinks after install is complete
```
mikemhenry, line 62:
```suggestion

```

.github/workflows/ci.yaml
```diff
@@ -93,11 +101,9 @@ jobs:
 
       working-directory: ./charmm
 
-    - name: CodeCov
-      if: ${{ github.repository == 'openmm/openmmforcefields'
-              && github.event != 'schedule' }}
+    - name: Upload coverage report to CodeCov
```
mikemhenry, line 104: I will add a quick guard to not upload coverage reports when it is scheduled CI

.gitignore
```diff
@@ -62,13 +65,16 @@ instance/
...
-docs/_build/
-
 # PyBuilder
 target/
 
+# PyCharm
+*.idea/vcs.xml
+
+# Jupyter
+=======
```
mikemhenry, line 75:
```suggestion

```

.github/workflows/ci.yaml
mikemhenry, line 62:
```suggestion

```

.gitignore
```diff
@@ -62,13 +65,16 @@ instance/
...
 # Jupyter Notebook
+>>>>>>> upstream/main
```
mikemhenry, line 77:
```suggestion

```

mikemhenry:
@mattwthompson I will take over from here! I just have a few things I want to tweak before I get it merged in, again, thanks for all your help!

mattwthompson:
Great!!! Thank you!

PR #229, opened 2022-08-30
Do not symlink data files in `setup.py`

mattwthompson:
With the deprecation of `python setup.py`, confusion (#124) and strange, difficult-to-diagnose behavior in CI (#225) I wanted to experiment with simplifying the distribution of data files by packaging them directly instead of creating symbolic links to them. Symlinks are generally not portable - may or may not help with the Windows issues. As far as I can tell, this requires no changes to the discovery machinery around `openmm.forcefielddir`.

This gets `python -m pip install .` working on my machine with both direct and editable installs.


mattwthompson:
Now moved to #231

PR #231, opened 2022-09-06
Do not symlink files in installation

mattwthompson:
#229 but not from a fork


jchodera:
Thanks! Will be great if this streamlines things!

PR #234, opened 2022-09-10
Turn off macos + Python 3.10 tests

mattwthompson:
This is a small change to see if the testing infrastructure is back to being online. If it is, everything should be green and codecov should (should! .... should!) report some useful information.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,9 +27,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


mattwthompson:
Well, that didn't work 🤷‍♂️

mikemhenry:
Not sure why the "test on macOS-latest..." are showing up, I've removed them from the list of tests that are required.

mikemhenry:
@mattwthompson I just removed all tests as "required" while we work on this to figure out which ones we actually want to require :)

PR #236, opened 2022-09-14
[DNM] #227 with codecov action

mattwthompson:
This switches back to `codecov/codecov-action` but is based off of a branch that should be passing.


mattwthompson:
@jchodera could you update the codecov settings to reflect the changing of the default branch? There should be a section about halfway down the page at https://app.codecov.io/gh/openmm/openmmforcefields/settings

PR #239, opened 2022-09-28
Convert all amber ions; add OPC and OPC3 waters

aizvorski:
This converts one-to-one all Amber ion frcmod files (from AmberTools 22) into OpenMM xml files with the same names and the same contents. Implements what is proposed in https://github.com/openmm/openmm/issues/3663  Moved here based on feedback from @peastman 

Run the conversion:

```
python ./amber/convert_amber_ions.py ~/anaconda3/pkgs/ambertools-22.0-py38h6177452_1
```

The new xml files end up in `amber/ffxml/ions/`

This does not change any of the existing ion conversion, and also does not combine water and ions in a single file.  The intention is to have the water+ions be made by using `<Include file="..."/>` mechanism in the xml, rather than by duplicating the contents; but that is a future change, at the moment this doesn't change any existing files.  It also doesn't use yaml - since the idea is to convert everything, there is nothing to configure.

Also, adds OPC and OPC3 waters, with the appropriate ion sets.  The waters are from https://github.com/openmm/openmm/pull/3654 and the opc_standard.xml file shows the proposed method for combining water and ions.

Feedback requested:
- How to test?  `validate_water_ion()` has a nice unit test for water+ions, but not obvious how to run that outside of a yaml recipe
- Any downsides to the suggested method of including a water and ions xml files to make the *standard.xml files?

PR #242, opened 2022-10-17
Convert all amber ions

mattwthompson:
This is #239 but not from a fork

```shell
$ git remote add aizvorski git@github.com:aizvorski/openmmforcefields.git
$ git fetch --all
$ git checkout aizvorski/convert-all-amber-ions
$ git checkout -b convert-all-amber-ions
$ git push upstream convert-all-amber-ions
```

Feel free to push to this branch if the original feature branch is updated.


mattwthompson:
I have `pytest openmmforcefields/tests/test_amber_import.py::test_ffxml_import -nauto` passing locally - we'll see if the runners agree

mattwthompson:
Okay, CI is passing now; there were some non-trivial changes brought in by merging upstream, so it might be worth another look. I'm not sure if the `amber/` folder should be blown away or not, the two XML files could be duplicates or unused.

I also can't comment on the meat of the original PR #239 - my objective here is to get that work closer to the finish line; feel free to push directly to this branch if more work is needed.

mikemhenry:
Would be good to have @aizvorski take a look at this

jchodera (approved):
This looks good to me!

Thanks so much for putting in the hard work to contribute this, @aizvorski and @mattwthompson!

aizvorski:
@mattwthompson Thank you very much for doing this!  That was a tough merge

A couple of renames and this should be good to go:

amber/opc.xml -> openmmforcefields/ffxml/amber/opc.xml
amber/opc3.xml -> openmmforcefields/ffxml/amber/opc3.xml

Could you possibly do that?  I'd do it but I'm not sure how :)

mattwthompson:
Let's see if that works

mattwthompson:
Okay, I think that worked. @aizvorski good to merge, do you think? Or @jchodera can merge when he gets to this next. I just don't want to do it without a final check.

aizvorski:
@mattwthompson Yep, good to merge!  Thanks again

Issue #247, opened 2022-11-22
Support custom openmm forces


jthorton:
When using a offxml with a [smirnoff-plugin](https://github.com/openforcefield/smirnoff-plugins) force such as the double exponential non-bonded potential which is implemented as a custom bond and non-bonded force in openmm they are dropped from the final system with no error. This seems to be due to this function which only preserves [class 1 forces](https://github.com/openmm/openmmforcefields/blob/d78f2b825888a68c17aee17420683a3816a4838f/openmmforcefields/generators/template_generators.py#L975). This also causes issues for downstream packages like QCEngine which uses the `SystemGenerator` class to build openmm systems. 

It would be great if custom forces could also be supported in someway or an error could be raise at least if some forces can not be converted.

mattwthompson:
I'm kinda curious how you custom forces get in at all? AFAICT the SMIRNOFF force fields are not loaded with plugins:

https://github.com/openmm/openmmforcefields/blob/d78f2b825888a68c17aee17420683a3816a4838f/openmmforcefields/generators/template_generators.py#L1282

jthorton:
Ahhh good catch, I forgot I patched my version of the toolkit to load plugins by default so I didn't have to keep changing other packages! So I think we can ignore the error raise part it seems to already be safe.

mattwthompson:
It could be that I haven't had my coffee yet, but I don't see how your custom forces can get into this and downstream packages without loading plugins? In other words, it seems like `openmmforcefields` should either load plugins by default or expose it as an option in the constructor.

jthorton:
Good point I think this has confused the issue I guess this should be a request to support plugins generally? I can remake it to make it more clear.

PR #248, opened 2022-11-22
Error if custom force is found

mattwthompson:
This is a short-term band-aid for #247 - implementing them will take more than a trivial amount of time so we should at least error out instead of silently ignoring entire forces in user's systems.

### DIFF
--- a/openmmforcefields/generators/template_generators.py
+++ b/openmmforcefields/generators/template_generators.py
@@ -1030,12 +1030,23 @@ class OpenMMSystemMixin:
                 element=element_symbol, mass=as_attrib(atom.mass))
             atom_type.set('class', atom.typename) # 'class' is a reserved Python keyword, so use alternative API
 
+        supported_forces = {
+            "NonbondedForce",
+            "HarmonicAngleForce",
+            "HarmonicBondForce",
+            "PeriodicTorsionForce",
+        }
+
         # Compile forces into a dict
         forces = dict()
         for force in system.getForces():
             force_name = force.__class__.__name__
+
             if force_name in forces:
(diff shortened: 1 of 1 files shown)


mikemhenry:
@mattwthompson @ijpulidos I closed a bunch of PRs that I think #290 took care of, is this one still needed?

openmmforcefields/generators/template_generators.py
```diff
@@ -1030,12 +1030,23 @@ def as_attrib(quantity):
...
         # Compile forces into a dict
         forces = dict()
         for force in system.getForces():
             force_name = force.__class__.__name__
+
             if force_name in forces:
-                raise Exception("Two instances of force {force_name} appear in System")
+                raise Exception(f"Two instances of force {force_name} appear in System")
+            if force_name not in supported_forces:
+                raise Exception(f"Custom forces not supported. Found force of type {force_name}.")
```
mattwthompson, line 1048: I should probably add I don't think this is the right exception to raise, I'm just not sure which it should be

mikemhenry: Eh, this package doesn't define many exceptions, but I made a new one

mikemhenry:
Not sure why this PR isn't running, @mattwthompson can you remake it when you have the chance?
image

mattwthompson:
I'm pretty sure https://github.com/openmm/openmmforcefields/issues/247 is unaddressed, it's not directly related to the other PRs we've had opened.

I'm not sure why I opened this on my fork, moving it to upstream now

mikemhenry:
Ah and yes it was a fork thing and not a permissions thing, oops

PR #264, opened 2023-04-18
Add support to load espaloma models trained using dropouts and batch normalization

### LINKED ISSUE #263: EspalomaTemplateGenerator cannot properly load espaloma models trained with dropouts and batch normalization
We need to call model.eval() after loading the espaloma model in [generators/template_generators.py#L1580](https://github.com/openmm/openmmforcefields/blob/main/openmmforcefields/generators/template_generators.py#L1580) when loading a model trained with dropouts and batch normalization. It will yield inconsistent inference results if model.eval() is not called.

Reference:
https://pytorch.org/tutorials/beginner/saving_loading_models.html

kntkb:
This PR fixes #263.


mikemhenry:
Let me remove the openeye license check since we don't really neeed openeye in this repo anyway

kntkb:
Closing PR because there is a newer PR #271 that supports this feature.

PR #271, opened 2023-05-22
Fix Espaloma system/template generator

### LINKED ISSUE #245: [Enhancement] Add options to EspalomaTemplateGenerator to deploy different espaloma models
The current implementation for EspalomaTemplateGenerator only accepts espaloma models that were jointly trained with charge losses which limits the use of different charge methods. Additionally, when the openmm system is created ([here](https://github.com/openmm/openmmforcefields/blob/main/openmmforcefields/generators/template_generators.py#1710)), there is no way to specify the forcefield used assign nonbonded parameters.

I think it would be great if we could have options to specify the charge method and the forcefield used to assign the nonbonded parameters.

kntkb:
This draft PR is fixes #245 and #263 by adding `template_generator_kwargs` to the system generators.

**ToDo**
- [ ] Add tests

### DIFF
--- a/openmmforcefields/generators/template_generators.py
+++ b/openmmforcefields/generators/template_generators.py
@@ -1569,13 +1569,13 @@ class EspalomaTemplateGenerator(SmallMoleculeTemplateGenerator,OpenMMSystemMixin
         self.espaloma_model_filepath = self._get_model_filepath(forcefield)
 
         # Check reference forcefield
-        installed_smirnoffs = SMIRNOFFTemplateGenerator().INSTALLED_FORCEFIELDS
         reference_forcefield = template_generator_kwargs['reference_forcefield']
-        if reference_forcefield not in installed_smirnoffs:
-            msg = f"Invalid reference forcefield. Choose from [{installed_smirnoffs}]."
+        try:
+            ff = ForceField("%s.offxml" % reference_forcefield)
+        except:
+            msg = f"Invalid reference forcefield. See https://github.com/openforcefield/openff-forcefields for supported force fields."
             ValueError(msg)
-        else:
-            self.reference_forcefield = reference_forcefield
+        self._reference_forcefield = reference_forcefield
 
         # Check charge method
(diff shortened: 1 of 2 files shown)


kntkb:
I think `EspalomaTemplateGenerator.generate_residue_template` will assign `nn` partial charges (espaloma charge) when the molecule is passed to `espaloma_model`. This means that partial charges assigned to a molecule will be overwritten. If we want to keep and use the original charges assigned to the molecule, we will need to book keep the information and reassign it after calling `espaloma_model`.

openmmforcefields/generators/template_generators.py
```diff
@@ -1705,14 +1726,16 @@ def generate_residue_template(self, molecule, residue_atoms=None):
         regenerate_impropers(molecule_graph)
 
         # Assign parameters
+        # NOTE: espaloma (nn) partial charges are assigned to molecules automatically if available.
+        # We need to overwrite the partial charges if we want to read them from the molecule.
```
ijpulidos, line 1730: Good catch! We need to make sure if this is the case because we want users to be able to use charges from other methods/models combined with espaloma. I'll raise an issue about this so we can keep track of this.

kntkb:
> I think `EspalomaTemplateGenerator.generate_residue_template` will assign `nn` partial charges (espaloma charge) when the molecule is passed to `espaloma_model`. This means that partial charges assigned to a molecule will be overwritten. If we want to keep and use the original charges assigned to the molecule, we will need to book keep the information and reassign it after calling `espaloma_model`.

This is now supported in https://github.com/openmm/openmmforcefields/pull/271/commits/c1987935f87d12436ff558be03b01c3ddf65c9cb

ijpulidos:
We need to make sure we respect this https://github.com/openmm/openmmforcefields/issues/282#issuecomment-1593845129

kntkb:
@ijpulidos Is it OK to close this PR since we have a #293 that replaces this purpose?

ijpulidos:
Yes, closing!

PR #287, opened 2023-07-11
Do not use old OpenFF Toolkit versions

mattwthompson:
`openff-toolkit <0.11` will no longer be supported soon; this drops the support and checks.


mikemhenry:
Superseded by #290

PR #288, opened 2023-07-11
Allow parsing of offxml strings

jthorton:
This PR allows an offxml string to be used as a small molecule force field for the `SystemGenerator` and `SMIRNOFFTemplateGenerator` see the example which now works with this PR.

```python
"Try and load a OFFXML string into a system generator object"
from openmmforcefields.generators import SystemGenerator
from openff.toolkit.typing.engines.smirnoff import ForceField
from openff.toolkit.topology import Molecule
import openmm

# load the ff
ff = ForceField("openff-2.0.0.offxml")
# create a system generator
system_gen = SystemGenerator(
    forcefields=['amber/ff14SB.xml', 'amber/tip3p_standard.xml'],
    small_molecule_forcefield=ff.to_string()
)
mol = Molecule.from_smiles("CC")
system = system_gen.create_system(topology=mol.to_topology().to_openmm(), molecules=mol)
with open("system.xml", "w") as output:
    output.write(openmm.XmlSerializer.serialize(system))
```

### DIFF
--- a/openmmforcefields/generators/template_generators.py
+++ b/openmmforcefields/generators/template_generators.py
@@ -1301,7 +1301,7 @@ class SMIRNOFFTemplateGenerator(SmallMoleculeTemplateGenerator,OpenMMSystemMixin
                 self._smirnoff_forcefield = openff.toolkit.typing.engines.smirnoff.ForceField(forcefield)
             except Exception as e:
                 _logger.error(e)
-                raise ValueError(f"Can't find specified SMIRNOFF force field ({forcefield}) in install paths") from e
+                raise ValueError(f"Can't find specified SMIRNOFF force field ({forcefield}) in install paths or parse the input as a string.") from e
 
         # Delete constraints, if present
         if 'Constraints' in self._smirnoff_forcefield._parameter_handlers:
--- a/openmmforcefields/generators/template_generators.py
+++ b/openmmforcefields/generators/template_generators.py
@@ -1275,14 +1275,19 @@ class SMIRNOFFTemplateGenerator(SmallMoleculeTemplateGenerator,OpenMMSystemMixin
 
         # Create ForceField object
         import openff.toolkit.typing.engines.smirnoff
-        try:
-            filename = forcefield
-            if not filename.endswith('.offxml'):
-                filename += '.offxml'
+
+        # check for an installed force field
+        available_force_fields = openff.toolkit.typing.engines.smirnoff.get_available_force_fields()
+        if (filename := forcefield + ".offxml") in available_force_fields or (filename := forcefield) in available_force_fields:
             self._smirnoff_forcefield = openff.toolkit.typing.engines.smirnoff.ForceField(filename)
-        except Exception as e:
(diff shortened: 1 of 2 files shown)


jthorton:
Confirmed this works when running the solvation leg of a benchmark free energy calculation via openfe.

mikemhenry:
@jthorton I did a merge commit to bring in a fix that was needed, let me know if it looks okay!

jthorton:
Thanks @mikemhenry looks like it changed the exception message slightly which is fine for me. Just want to check if my use of the walrus `:=` was okay? I think this will only work with python 3.8+

mikemhenry:
I am happy with using python 3.8+ language features. We can even use python 3.9+ language features since we follow NEP29 (or try to)

mattwthompson:
OpenFF is pretty good about using it, so congrats - you've inherited it!

mikemhenry:
Ah classic, can't run this from a fork. @jthorton I will push this to a branch in the repo so we can make sure it passes in CI

mikemhenry:
@jthorton I don't know if this trick will keep working if you make a commit, so ping me if you do and CI stops running

mikemhenry:
Looks like it passes :rocket: @ijpulidos would you mind reviewing this one? Should we wait for #293 first?

mikemhenry (approved):

mikemhenry:
I don't think we need to wait on any other PRs, just need another review from @ijpulidos

openmmforcefields/generators/template_generators.py
```diff
@@ -1289,14 +1289,19 @@ def __init__(self, molecules=None, cache=None, forcefield=None):
...
-            _logger.error(e)
-            raise ValueError(f"Can't find specified SMIRNOFF force field ({forcefield}) in install paths") from e
+
+        # just try parsing the input and let openff handle the error
+        else:
+            try:
+                self._smirnoff_forcefield = openff.toolkit.typing.engines.smirnoff.ForceField(forcefield)
+            except Exception as e:
+                _logger.error(e)
+                raise ValueError(f"Can't find specified SMIRNOFF force field ({forcefield}) in install paths") from e
```
ijpulidos, line 1304: I think we want to improve the error message here, since as far as I can see it's not now that we don't find the ff on the install paths but it's rather that the provided string is not correctly formatted as a ffxml string.

Is it too crazy to add some more logic here such that we can have specific error messages as to what is causing this? It doesn't seem too hard to accomplish, as far as I can tell.

ijpulidos (commented):
This is looking great! I just added some comment about the error message when you get a poorly formatted string.

openmmforcefields/generators/template_generators.py
jthorton: I have updated the error to reflect that a missing file or poorly formatted string could cause this error. As for specific messages on the formatting of the forcefield I would prefer to leave this to the `openff.toolkit.ForceField` and just raise the error from there as it's quite descriptive itself.

Issue #280, opened 2023-06-14
Update to AmberTools23


jchodera:
AmberTools23 includes a new GAFF version (2.2.20, March 2021). From the header:
```
AMBER General Force Field for organic molecules (Version 2.2.20, March 2021)
```
We should add this GAFF version and update other force fields we are able to.

mattwthompson:
The short version of the current plan is

1. Make a 0.13.0 that doesn't have GAFF support (#329)
  a. Point users towards 0.12.0 which still does, with limitations (Python 3.10, AmberTools 22, x86 architecture only)
  a. These limitations don't apply to the SMIRNOFF template generator
3. Make a 0.14.0 that relies on https://github.com/openmm/openmm/pull/4531
  a. This should work with Python 3.10-3.12, AmberTools 22-23, x64 and Apple Silicon
4. When AmberTools 24 is packaged, test this package alongside it
  b. Currently expect it to work (no new GAFF version) but hard to predict prior to testing

mattwthompson:
I think this is done? @mikemhenry

epretti:
As of #368, the latest GAFF 2.2.20 from AmberTools 24.8 is in, so it looks like this can be closed.  Feel free to reopen if something remains outstanding here.

PR #289, opened 2023-07-11
[DNM] Pinning ambertools version to <= 22

### LINKED ISSUE #286: Pin to ambertools <= 22 until we can sort out changes to GAFF2 behavior
cc : #281

ijpulidos:
Ambertools 23 changed the behavior and it is causing failures with our tests. This pins the version for now so we can get a release while we find a solution for #280 

Resolves #286

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,7 +28,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: ["3.9", "3.10"]
+        python-version: ["3.9", "3.10", "3.11"]
         latest-openff-toolkit: [true, false]
         exclude:
           - python-version: "3.10"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,7 +28,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.9", "3.10"]
         latest-openff-toolkit: [true, false]
         exclude:
           - python-version: "3.10"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,7 +28,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: ["3.9", "3.10", "3.11"]
+        python-version: ["3.9", "3.10"]
         latest-openff-toolkit: [true, false]
         exclude:
           - python-version: "3.10"
--- a/devtools/conda-envs/test_env.yaml
+++ b/devtools/conda-envs/test_env.yaml
@@ -24,7 +24,7 @@ dependencies:
 
     # Requirements for conversion tools
   - pyyaml
-  - ambertools >=18.0 # contains sufficiently recent ParmEd
+  - ambertools <=22.0 # contains sufficiently recent ParmEd
   - lxml
   - networkx
 
--- a/devtools/conda-envs/test_env.yaml
+++ b/devtools/conda-envs/test_env.yaml
@@ -41,6 +41,6 @@ dependencies:
     # TODO: Rework this once espaloma is on conda-forge
     #
   - pytorch >=1.8.0
-  - dgl
+  - dgl <1.1
   - qcportal >=0.15.0
   - espaloma
--- a/devtools/conda-envs/test_env.yaml
+++ b/devtools/conda-envs/test_env.yaml
@@ -24,7 +24,7 @@ dependencies:
 
     # Requirements for conversion tools
   - pyyaml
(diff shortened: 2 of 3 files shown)


ijpulidos:
I'll also be dropping testing with python 3.8 in our CI in this same PR.

mattwthompson:
This failure

```
FAILED openmmforcefields/tests/test_template_generators.py::TestSMIRNOFFTemplateGenerator::test_energies - ValueError: Can't find specified SMIRNOFF force field (opc3) in install paths
```
stems from OPC being shipped by `openff-forcefields`. You can cherry-pick a band-aid from here: https://github.com/openmm/openmmforcefields/pull/287/commits/4d7c50ff840a870ec98bc3300ada98c7feaaeaa2 or just leave it for #287

devtools/conda-envs/test_env.yaml
```diff
@@ -24,7 +24,7 @@ dependencies:
 
     # Requirements for conversion tools
   - pyyaml
-  - ambertools >=18.0 # contains sufficiently recent ParmEd
+  - ambertools <=22.0 # contains sufficiently recent ParmEd
```
mattwthompson, line 27: You might want to avoid solvers looking into silly old versions (18, etc.) by adding a lower constraint:

```suggestion
  - ambertools >=20,<=22.0
```

or, what I'd do:

```suggestion
  - ambertools =22.0
```

.github/workflows/ci.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.9", "3.10"]
```
mattwthompson, line 31: 3.11 probably works, not necessarily advocating for it to be added here

mattwthompson (approved):

ijpulidos:
@mattwthompson Thanks! I think resolving #287 before any other PR makes sense, unless folks have some objection about it. 

Any ideas about this one https://github.com/openmm/openmmforcefields/actions/runs/5524287757/jobs/10076389477#step:9:752 ? I'm clueless about that one

devtools/conda-envs/test_env.yaml
ijpulidos: Yeah, I was thinking about it, I think it's best if we pin it to 22. THanks!

.github/workflows/ci.yaml
ijpulidos: Can't hurt, thanks!

mattwthompson:
> Any ideas about this one

The error is confusing, but it's actually the same issue as when water models are included in `SMIRNOFFTemplateGenerator.INSTALLED_FORCEFIELDS` or whatever that variable is called. The water models don't include AM1-BCC, just library charges for water, so the simplest of molecules fail charge assignment. The error is not useful (there should be a way to capture which force field failed and bubble that up, I just haven't done the work) and I think it might have been @mikemhenry who did the detective work somewhere else in this repo a few months ago.

ijpulidos:
@mattwthompson As far as I can tell it's happening with the `opc3` force field and the `C=O` molecule, as per https://github.com/openmm/openmmforcefields/actions/runs/5524287757/jobs/10076389477#step:9:601

ijpulidos:
So I guess it doesn't make sense to parametrize a non-water molecule with the `opc3` force field.

mikemhenry:
https://github.com/openmm/openmmforcefields/pull/273

mattwthompson:
Right - I'm suggesting that `openmmforcefields` does not include water models in its list of small molecule force fields at all. Probably having an allow list with `openff-[0-9].[0-9].[0-9].offxml` type of regex would be a safer approach when collecting force fields

.github/workflows/ci.yaml
ijpulidos: 3.11 is incompatible with ambertools 22, skipping that one for now.

mikemhenry:
Okay that isn't a bad idea, but will there be an issue if a user supplies their own SMIRNOFF xml (from like bespoke fit)?

mattwthompson:
That's fair - #288 would be a workaround but not the most elegant choice. You could instead just filter out OPC/TIP4P/etc. in the test, since I think it just naively runs through all available force fields whether or not it makes sense to try them on a molecule. That would prevent restrictions on what gets fed through the `get_available_force_fields` machinery.

If a user submits a bug report that TIP4P isn't parametrizing aspirin, I think we'd close that hypothetical issue as invalid, after all.

Anyway, you'll probably just want to go for whatever gets CI passing again and most promptly, and I'll update #287 accordingly.

mikemhenry:
Ah I misunderstood the issue, @ijpulidos yes feel free to just setup some logic to skip for this test (I thought that work around would affect the `get_available_force_fields` machinery) and we could clean up the logic of `get_available_force_fields` in a different PR

ijpulidos:
This one is superseded by #290 , I'm going to have this for testing purposes for now.

mikemhenry:
Superseded by #290

PR #290, opened 2023-07-12
Scattered fixes to get CI running

mattwthompson:
* Pin AmberTools
* Skip non-small molecule force fields at test time
* Always run tests, even if OpenEye is not licensed
* Update supported Python versions


openmmforcefields/tests/test_template_generators.py
```diff
@@ -780,6 +780,11 @@ def test_energies(self):
         for small_molecule_forcefield in SMIRNOFFTemplateGenerator.INSTALLED_FORCEFIELDS:
             if "ff14sb" in small_molecule_forcefield:
                 continue
+            if "tip" in small_molecule_forcefield:
+                continue
+            if "tip" in small_molecule_forcefield:
```
ijpulidos, line 785: Did you mean `opc` here?

mattwthompson: yup

mattwthompson:
This inheritance is having some side effects - could it be updated to only pull from the base class?

https://github.com/openmm/openmmforcefields/blob/87226a280862558c16b32b9b9e59eea0fb3bcd14/openmmforcefields/tests/test_template_generators.py#L710

Removing more of the `unittest` reliance might be nice, especially since the developers here spend a good bit of time digging through test failures.

mattwthompson:
This PR gets everything working except DGL stuff - could somebody take this over?

jchodera:
Thanks so much, @mattwthompson ! 

@ijpulidos will take this over!

ijpulidos:
I changed the tests such that we can test espaloma independently of the other forcefields. Tests seem to be passing now (still a few to go, but I already see green check marks). The coverage of course was reduced because now we are skipping the espaloma parts and I think it's working as expected as per https://app.codecov.io/gh/openmm/openmmforcefields/pull/290/blob/openmmforcefields/generators/template_generators.py , but would love somebody else to confirm this.

I marked the espaloma tests with `pytest.mark.espaloma` but for some reason they are not being run correctly in the Espaloma CI workflow (check https://github.com/openmm/openmmforcefields/actions/runs/5581424990/jobs/10199559997?pr=290#step:6:24) . I need to figure out what else is needed for these to get run in that workflow exclusively. 

The good news is that we seem to have the CI back up and running succesfully again for all the non-espaloma ffs that we are testing.

ijpulidos:
I see the tests are failing about the `opc3` forcefields with the old `openff-toolkit`, I think this is expected. @mattwthompson do you have any insights on this? Should we even be testing 0.10.x versions of the toolkit?

mattwthompson:
I would just drop the 0.10.x toolkit testing, support is completely going away soon and as far as I'm aware all projects have upgraded to at least 0.11.x at this point. If people need it with `openmmforcefields`, they can install an old version, and if they need a bugfix release they can advocate for it.

You can try to `git cherry-pick ac9881e092b1bdecfe9167b5c3cf469281833387` and then clean up my mess, or just crib from #287 whatever you want.

mikemhenry:
I think you need to add the path to the conftest.py file at the end of this command ` pytest -v --log-cli-level $LOGLEVEL $COV_ARGS --durations=20 \
    -k "TestEspalomaTemplateGenerator" -m "espaloma"`

devtools/conda-envs/test_env.yaml
```diff
@@ -41,6 +41,6 @@ dependencies:
     # TODO: Rework this once espaloma is on conda-forge
     #
   - pytorch >=1.8.0
-  - dgl
+  - dgl <1.1
```
mikemhenry, line 44:
```suggestion
  - dgl <1
```

mikemhenry (approved):

PR #292, opened 2023-07-19
Error on custom force

mattwthompson:
See #248


mikemhenry:
This doesn't fix the issue entirely but provides better error messages

Issue #296, opened 2023-08-16
Make passing user supplied partial charges to SystemGenerator an explicit keyword?


IAlibay:
Currently SystemGenerator does the following:
- If there are partial charges present on the OFF molecule use them directly (i.e. don't generate new charges)
- Unless these partial charges are zeroed, in which case silently re-generate them.

From recent discussions (alchemiscale meeting ~ month ago) it seems like folks are a bit wary about the idea of this being done silently. Indeed the latter behaviour caught me out whilst I was trying to debug something else lately.

Would folks be interested in either:
- Making using pre-defined charges something you need to pass an explicit kwarg for?
- Being rather noisy about what's going on (something higher level than logger.debug)?

ijpulidos:
I think this is something that have been somewhat discussed in the past in https://github.com/openmm/openmmforcefields/issues/282#issuecomment-1593845129 The concern here is that it would break the API

I'm okay with either approach, but I agree that we should probably make this clearer in the logging for users.

PR #300, opened 2023-09-05
Use gasteiger instead of am1bcc for large mols

jaketanderson:
This small edit makes toolkit assign gasteiger partial charges instead of am1bcc if the molecule's atom count is >= 150, for speed purposes.

### DIFF
--- a/openmmforcefields/generators/template_generators.py
+++ b/openmmforcefields/generators/template_generators.py
@@ -607,11 +607,16 @@ class GAFFTemplateGenerator(SmallMoleculeTemplateGenerator):
         if self._molecule_has_user_charges(molecule):
             _logger.debug(f'Using user-provided charges because partial charges are nonzero...')
         else:
-            _logger.debug(f'Computing AM1-BCC charges...')
+            _logger.debug(f'Computing partial charges...')
             # NOTE: generate_conformers seems to be required for some molecules
             # https://github.com/openforcefield/openff-toolkit/issues/492
             molecule.generate_conformers(n_conformers=10)
-            molecule.assign_partial_charges(partial_charge_method='am1bcc')
+            if len(molecule.atoms) >= 150:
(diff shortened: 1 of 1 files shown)


mattwthompson:
Looks like this is only for the GAFF template generator, correct? I'm not familiar enough with GAFF to judge whether or not this is canonical, though I do understand why one would want this.

jaketanderson:
Yeah I think partial charge assignment via toolkit only happens for the GAFF generator. The supposition here is that as the molecule grows large, the computational cost of AM1BCC grows faster than does the rate of mistypings made using Gasteiger. This is just my intuition though...

j-wags:
In its current form I'd consider this unwanted behavior, since it will change the meaning of "GAFF" and many users would miss the warning and publish flawed studies. I'm not a maintainer, but as a community member I'd think this should only be triggered if the user sets a kwarg like `use_gasteiger_for_large_molecules=True`.

mattwthompson:
In that case #296 and letting users to do this themselves would be a simpler fix

jaketanderson:
@j-wags that is a valid concern, I appreciate you saying that. If I'm the only person really facing this problem then I'm fine with editing my local `openmmforcefields` package. I don't want to add confusion for everyone else at the sake of my own convenience :)

j-wags:
@jaketanderson Thanks for understanding and working with the changes locally. And I agree that #296 would be a more graceful solution for this!

mattwthompson:
In this case I think I have to be the bad guy and close this - hopefully the `charge_from_molecules` pathway can be enabled soon!

AM1-BCC for large molecules is a common pitfall for users, so if this is enabled in the future it would make a nice example. It shouldn't take too much code when supported: assign charges using the `Molecule` API and show how `charge_from_molecules` can be fed through the GAFF template generator to bypass AM1-BCC.

PR #305, opened 2023-10-19
Update SMIRNOFF/OpenFF docs/examples

mattwthompson:
I noticed that the README currently recommends an old (2020?) Parsley force field. I updated the default behavior of `SMIRNOFFTemplateGenerator` to use 2.1.0, our current recommended force field for general use. That should be the only behavior change introduced here.

I also took a dive through the examples in the README and many of the examples in the source code and updated them and/or made sure they ran. I don't think there are substantive or wacky changes there.

### DIFF
--- a/openmmforcefields/generators/template_generators.py
+++ b/openmmforcefields/generators/template_generators.py
@@ -46,7 +46,7 @@ class SmallMoleculeTemplateGenerator:
 
         Parameters
         ----------
-        molecules : openff.toolkit.topology.Molecule or list, optional, default=None
+        molecules : openff.toolkit.Molecule or list, optional, default=None
             Can alternatively be an object (such as an OpenEye OEMol or RDKit Mol or SMILES string) that can be used to construct a Molecule.
             Can also be a list of Molecule objects or objects that can be used to construct a Molecule.
             If specified, these molecules will be recognized and parameterized as needed.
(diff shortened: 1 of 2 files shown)


openmmforcefields/generators/template_generators.py
```diff
@@ -1603,6 +1603,7 @@ def __init__(self, molecules=None, cache=None, forcefield=None, model_cache_path
             self._reference_forcefield = template_generator_kwargs.get('reference_forcefield', 'openff_unconstrained-2.0.0')
             self._charge_method = template_generator_kwargs.get('charge_method', 'nn')
         else:
+            # Consider upgrading to 2.1.0, the recommended small moleucle force field for general use
             self._reference_forcefield = 'openff_unconstrained-2.0.0'
             self._charge_method = 'from-molecule'
```
mattwthompson, line 1608: I noticed this in `EspalomaTemplateGenerator` but didn't want to make the change without better understanding what would happen

jchodera: I believe the LJ parameters are unchanged between 2.0.0 and 2.1.0, so this should have no effect. But thanks for adding the comment for us to reevaluate later!

jchodera (approved):
This is great! Thanks!

openmmforcefields/generators/template_generators.py
mattwthompson: I think that's right - it's a Pavlovian response of mine 🙂

PR #320, opened 2023-12-20
14scaling lj coul template generators

mikemhenry:
just a trick to get CI to work on a fork


mikemhenry:
I don't actually need to open the PR for the trick to work TIL

PR #332, opened 2024-05-03
linted all files

mikemhenry:


mattwthompson:
I ran into some frustration with the docstrings being long in line length (which isn't inherently a horrible thing IMO) and flake8 or ruff not being happy

openmmforcefields/tests/test_template_generators.py
```diff
@@ -1,3 +1,4 @@
+# ruff: noqa: E501
```
mikemhenry, line 1: @mattwthompson yes I don't think it a bad thing, so I just had it skip this whole file and this was the only file where they were too long

mattwthompson: Seems good, the config is still pretty simple

PR #344, opened 2024-07-22
Remove shipping gaff/ffxml and read parms from parmchk2

mikemhenry:


mikemhenry:
the atom types are not actually duplicated in the xml https://github.com/openmm/openmmforcefields/actions/runs/10047847083/job/27770503705?pr=344#step:6:1541

I think the issue comes from loading multiple forcefield xmls with the same atom type here: https://github.com/openmm/openmmforcefields/blob/main/openmmforcefields/generators/template_generators.py#L350

I am a bit confused with how this used to work with the old parmchk2 output/how we made sure we didn't load duplicate atom types

mikemhenry:
I am pretty sure we will need to add a check in openmmforcefields somewhere to prevent registering duplicate atom types since if we call `forcefield.loadFile(StringIO(ffxml_contents))` for each molecule we generate the parameters for, we will end up with duplicate atom types that `openmm < 8.1.1` will complain about. I still think this is an improvement (where we call parmchk2 and parse the output to get the parameters every time) but we need to add a bit of logic where we keep track of the atom types we have already loaded in

openmmforcefields/generators/template_generators.py
```diff
@@ -347,7 +347,11 @@ def generator(self, forcefield, residue):
                         outfile.write(ffxml_contents)
 
                 # Add the parameters and residue definition
-                forcefield.loadFile(StringIO(ffxml_contents))
+                try:
+                    forcefield.loadFile(StringIO(ffxml_contents))
+                except (ValueError, KeyError) as err:
+                    print(ffxml_contents)
+                    raise err
```
mikemhenry, line 354: this will get un-done and was just to make debugging easier

amber/convert_amber.py
```diff
@@ -21,7 +21,7 @@
 import yaml
 from lxml import etree as et
 from parmed.exceptions import ParameterWarning
-from pkg_resources import resource_filename
+import importlib_resources
```
mikemhenry, line 24: I can make thse changes in another PR, but `pkg_resources` is deprecated

devtools/conda-envs/test_env.yaml
```diff
@@ -8,11 +8,11 @@ dependencies:
   - pytest-cov
   - pytest-xdist
   - pytest-randomly
-  - openmm >=8.1.2,<9
+  - openmm <8.1.1
```
mikemhenry, line 11: this is just to show that this fix works with older OpenMM versions

openmmforcefields/generators/template_generators.py
```diff
@@ -594,11 +607,11 @@ def generator(self, forcefield, residue):
...
-        if forcefield not in self._gaff_parameters_loaded:
-            # Instruct the ForceField to load the GAFF parameters
-            forcefield.loadFile(self.gaff_xml_filename)
-            # Note that we've loaded the GAFF parameters
-            self._gaff_parameters_loaded[forcefield] = True
+        # if forcefield not in self._gaff_parameters_loaded:
+        #    # Instruct the ForceField to load the GAFF parameters
+        #    forcefield.loadFile(self.gaff_xml_filename)
+        #    # Note that we've loaded the GAFF parameters
+        #    self._gaff_parameters_loaded[forcefield] = True
```
mikemhenry, line 614: We don't need to subclass anymore since we don't load in the gaff xmls -- I will remove it but for testing I just commented it out for now

openmmforcefields/generators/template_generators.py
```diff
@@ -730,6 +743,20 @@ def generate_residue_template(self, molecule, residue_atoms=None):
...
+        # remove and add them as needed
+        # see if this fixes it
+        # params.atom_types.keys()
+        for atom_type in params.atom_types.copy().keys():
+            if atom_type not in self._gaff_atom_type_cache:
+                self._gaff_atom_type_cache.add(atom_type)
+                print(f"added {atom_type} to cache")
+            else:
+                del params.atom_types[atom_type]
+                print(f"del {atom_type}")
```
mikemhenry, line 759: I will clean this up and explain it with doc strings, I just wanted to get this tested on CI to make sure it worked (not just locally)

but the tl;dr is we need this or else we end up trying to register duplicate atom types with openmm as we loop over molecules and generate parameters

openmmforcefields/tests/conftest.py
```diff
@@ -21,6 +21,7 @@ def pytest_addoption(parser):
 
 def pytest_configure(config):
     config.addinivalue_line("markers", "espaloma: mark test as slow to run")
+    config.addinivalue_line("markers", "gaff: mark test as using gaff")
```
mikemhenry, line 24: this fixes warnings about pytest not having the `gaff` mark registered -- I can also move this to a different PR

openmmforcefields/ffxml/amber/gaff/ffxml/gaff-1.4.xml
mikemhenry, line 1: Should we keep these but just not use them? I could see someone using them for _reasons_. I was thinking we delete them since we don't _need_ them but we don't need to delete them.

openmmforcefields/generators/template_generators.py
```diff
@@ -878,7 +905,8 @@ def read_file_contents(filename):
 
             # Run parmchk.
             shutil.copy(self.gaff_dat_filename, "gaff.dat")
-            cmd = f"parmchk2 -i out.mol2 -f mol2 -p gaff.dat -o out.frcmod -s {self._gaff_major_version}"
+            cmd = f"parmchk2 -i out.mol2 -f mol2 -p gaff.dat -o out.frcmod -s {self._gaff_major_version} -a Y"
```
mikemhenry, line 908: This is the most important part, here we are telling `parmchk2` to print out all of the parameters. We used to print out only parms that didn't exist in the gaff.dat file. What this enables:

1) we don't have to convert new gaff.dat versions to xmls
2) we can support whatever gaff versions that `parmchk2` supports (from the given ambertools version)

openmmforcefields/generators/template_generators.py
mikemhenry: btw, we didn't need with the old way of doing things since older versions of ambertools/parmchk2 didn't print out atom types that existed in `gaff.dat` with the way we ran parmchk2 (new versions of parmchk2 started adding atom type `c5` to the frcmod file)

mikemhenry:
Awesome! CI is green - going to clean this up now and request reviews

openmmforcefields/generators/template_generators.py
```diff
@@ -348,6 +348,7 @@ def generator(self, forcefield, residue):
 
                 # Add the parameters and residue definition
                 forcefield.loadFile(StringIO(ffxml_contents))
+
```
mikemhenry, line 351:
```suggestion
```

mikemhenry:
@jchodera @mattwthompson ready for review!

The plan is after we merge this, we can cut another release of `openmmforcefields` and relax the openmm version on our conda-forge package.

PR #359, opened 2025-02-03
Update macOS and OpenMM versions in CI

epretti:
`macos-12` is deprecated and as of now is completely unsupported causing those CI runs to fail: see <https://github.com/actions/runner-images/issues/10721>.

This also updates the OpenMM versions to the latest two releases.  I am assuming that this is the convention used; please correct if not.


mattwthompson:
`macos-13` is also going to give you problems FYI https://github.com/conda-forge/ambertools-feedstock/issues/140

PR #361, opened 2025-02-07
Test new ambertools 24  build

mikemhenry:
First I want to make sure CI is green, then I will see if tests pass installing `ambertools` from outside `conda-forge`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, macos-13]
+        os: [ubuntu-latest, macos-latest]
         python-version: ["3.10", "3.11", "3.12"]
         amber-conversion: [false]
         charmm-conversion: [false]
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-latest, macos-13]
         python-version: ["3.10", "3.11", "3.12"]
         amber-conversion: [false]
         charmm-conversion: [false]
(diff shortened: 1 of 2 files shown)


mikemhenry:
Everything passed except the `macos-12` runners, which are depreciated. `macos-13` and `macos-14` (aka `macos-latest`) are the only supported ones (`macos-15` is in beta). See https://github.com/actions/runner-images

mikemhenry:
Okay, now going to try pinning to the older build to see if that fixes macos-13

mikemhenry:
okay cool, looks like no regressions with ambertools 24 at least -- same issue on intel macs but that is okay. we should add openmm 8.2 to the testing matrix to double check and then I think we can update the feedstock to work with ambertools 24

mikemhenry:
Ignore the missing "Required" checks -- they are old mac versions and openmm versions that we don't need to test anymore IMHO

mikemhenry:
@epretti 
Does this look okay? We now test against the latest ambertools + the last two versions of openmm + the latest osx (arm64) and linux

epretti:
Great, thanks for getting this working!

And if I am understanding correctly, this means that there is now a working package for AmberTools 24 that is compatible with openmmforcefields, so we shouldn't need to drop AmberTools as a dependency anymore?  If so, I can close conda-forge/openmmforcefields-feedstock#19.

mikemhenry:
@epretti Yes! -- I am wating to merge in https://github.com/conda-forge/openmmforcefields-feedstock/pull/20 until we can get a env with python 3.13 and ambertools24 + openff-toolkit -- right now it isn't working since there is a libboost/pydantic/openff-toolkit log jam BUT for python version 3.10, 3.11, and 3.12 -- everything is working!

mikemhenry:
@mattwthompson Thoughts on merging this in and then dropping the required macos-12 checks and adding the new ones as required?

mattwthompson:
+1 on dropping `macos-12`
+1 on dropping AmberTools 23 if version 24 is the only one we trust

mattwthompson (approved):

PR #367, opened 2025-02-27
Remove CMMRemover (if present) in SMIRNOFFTemplateGenerator

### LINKED ISSUE #365: Interchange 0.4.2 breaks SMIRNOFF template generator
It looks like the well-intentioned convenience behavior added in https://github.com/openforcefield/openff-interchange/pull/1133 has [broken](https://github.com/openmm/openmmforcefields/actions/runs/13557150840/job/37893557675#step:6:2840) the SMIRNOFFTemplateGenerator. 

(thanks @dotsdl for initially spotting this)

For folks getting this error on freshly made envs, you can get around this by downpinning to `openff-interchange=0.4.1`

j-wags:
- [x] Fixes #365

### DIFF
--- a/openmmforcefields/generators/template_generators.py
+++ b/openmmforcefields/generators/template_generators.py
@@ -1570,6 +1570,7 @@ class SMIRNOFFTemplateGenerator(SmallMoleculeTemplateGenerator, OpenMMSystemMixi
         * Atom names in molecules will be assigned Tripos atom names if any are blank or not unique.
 
         """
+        from openmm import CMMotionRemover
         # Use the canonical isomeric SMILES to uniquely name the template
         smiles = molecule.to_smiles()
         _logger.info(f"Generating a residue template for {smiles} using {self._forcefield}")
--- a/openmmforcefields/generators/template_generators.py
+++ b/openmmforcefields/generators/template_generators.py
@@ -1592,9 +1592,9 @@ class SMIRNOFFTemplateGenerator(SmallMoleculeTemplateGenerator, OpenMMSystemMixi
 
         # Remove CMMotionRemover if present
         # See https://github.com/openmm/openmmforcefields/issues/365
-        for f_idx in range(system.getNumForces()):
+        for f_idx in reversed(range(system.getNumForces())):
             force = system.getForce(f_idx)
-            if "CMMotionRemover" in str(force):
+            if isinstance(force, CMMotionRemover):
                 system.removeForce(f_idx)
 
         self.cache_system(smiles, system)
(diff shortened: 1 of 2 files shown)


openmmforcefields/generators/template_generators.py
```diff
@@ -1589,6 +1589,14 @@ def generate_residue_template(self, molecule, residue_atoms=None):
         system = self._smirnoff_forcefield.create_openmm_system(
             molecule.to_topology(), charge_from_molecules=charge_from_molecules
         )
+
+        # Remove CMMotionRemover if present
+        # See https://github.com/openmm/openmmforcefields/issues/365
+        for f_idx in range(system.getNumForces()):
+            force = system.getForce(f_idx)
+            if "CMMotionRemover" in str(force):
+                system.removeForce(f_idx)
```
mattwthompson, line 1598: A little safer - hard to expect multiple CMM removers, but this is modifying a list while iterating over it. Not sure what happens if, for example, the force is listed last

```suggestion
                break
        system.removeForce(f_idx)
```

mattwthompson (approved):

openmmforcefields/generators/template_generators.py
peastman: Loop in reverse order from `system.getNumForces()-1` down to 0.  That way the indices will still be correct if you remove one.

The best way to check the force type is `if isinstance(force, CMMotionRemover)`.

Why are you removing CMMotionRemovers?

mattwthompson: The check here flags it as a custom force: https://github.com/openmm/openmmforcefields/blob/7a32d2d109a6cb44dd4344ef0482d7cf7f352cdd/openmmforcefields/generators/template_generators.py#L1116-L1131

peastman: That means you should fix the check, not remove the force.  The default for `createSystem()` is `removeCMMotion=True`.  This change would break it.

mattwthompson: This check happens when generating the residue template from OpenFF tooling, not the system that's ultimately passed to the user

mikemhenry: I would need to look at the code more closely, but what happens if when iterating over `system.getForces()` you just `continue` if the force is `CMMotionRemover`? That should keep it out of the `forces` dict?

mikemhenry: Okay so if we go with my idea, `CMMotionRemover` won't end up in the `ffxml`  when you call `generate_residue_template` -- is that a problem? If it is, then all we need to do is write `CMMotionRemover` into the `ffxml`

mattwthompson: Strong preference for 1

epretti (approved):
Looks good to me.  Since Matt has also approved, I will go ahead and merge.

PR #370, opened 2025-03-21
Improve molecule filtering in tests

### LINKED ISSUE #330: SMIRNOFF energy test failing
While working on something else, haven't looked any deeper
```
___________________________________ TestSMIRNOFFTemplateGenerator.test_energies ___________________________________

self = <openmmforcefields.tests.test_template_generators.TestSMIRNOFFTemplateGenerator testMethod=test_energies>

    def test_energies(self):
        """Test potential energies match between openff-toolkit and OpenMM ForceField"""

        # Test all supported SMIRNOFF force fields
        for small_molecule_forcefield in SMIRNOFFTemplateGenerator.INSTALLED_FORCEFIELDS:
            if "ff14sb" in small_molecule_forcefield:
                continue
            if "tip" in small_molecule_forcefield:
                continue
            if "opc" in small_molecule_forcefield:
                continue

            # We cannot test openff-2.0.0-rc.1 because it triggers an openmm.OpenMMException due to an equilibrium angle > \pi
            # See https://github.com/openmm/openmm/issues/3185
            if "openff-2.0.0-rc.1" not in small_molecule_forcefield:
                continue

            print(f'Testing energies for {small_molecule_forcefield}...')
            # Create a generator that knows about a few molecules
            # TODO: Should the generator also load the appropriate force field files into the ForceField object?
            generator = SMIRNOFFTemplateGenerator(molecules=self.molecules, forcefield=small_molecule_forcefield)
            # Create a ForceField
            openmm_forcefield = openmm.app.ForceField()
            # Register the template generator
            openmm_forcefield.registerTemplateGenerator(generator.generator)
            # Parameterize some molecules
            for molecule in self.molecules:
                # Create OpenMM System using OpenMM app
                openmm_system = openmm_forcefield.createSystem(molecule.to_topology().to_openmm(), removeCMMotion=False, nonbondedMethod=NoCutoff)

                # Retrieve System generated by the SMIRNOFF typing engine
                smirnoff_system = generator.get_openmm_system(molecule)

                # Compare energies and forces
>               self.compare_energies(molecule, openmm_system, smirnoff_system)

openmmforcefields/tests/test_template_generators.py:819:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
openmmforcefields/tests/test_template_generators.py:647: in compare_energies
    reference_energy, reference_forces = cls.compute_energy(
openmmforcefields/tests/test_template_generators.py:612: in compute_energy
    context = openmm.Context(system, integrator, platform)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <openmm.openmm.Context;  >
args = (<openmm.openmm.System; proxy of <Swig Object of type 'OpenMM::System *' at 0x170762b50> >, <openmm.openmm.VerletInteg...or *' at 0x170761560> >, <openmm.openmm.Platform; proxy of <Swig Object of type 'OpenMM::Platform *' at 0x170762d00> >)

    def __init__(self, *args):
        r"""
        __init__(self, system, integrator) -> Context
        __init__(self, system, integrator, platform) -> Context
        __init__(self, system, integrator, platform, properties) -> Context
        __init__(self, other) -> Context
        Construct a new Context in which to run a simulation, explicitly specifying what Platform should be used to perform calculations and the values of platform-specific properties.

        Parameters
        ----------
        system : System
            the System which will be simulated
        integrator : Integrator
            the Integrator which will be used to simulate the System
        platform : Platform
            the Platform to use for calculations
        properties : map< std::string, std::string >
            a set of values for platform-specific properties. Keys are the property names.
        """
>       _openmm.Context_swiginit(self, _openmm.new_Context(*args))
E       openmm.OpenMMException: HarmonicAngleForce: angle must be between 0 and pi

../../micromamba/envs/interchange-examples-env/lib/python3.11/site-packages/openmm/openmm.py:2471: OpenMMException
---------------------------------------------- Captured stdout call -----------------------------------------------
Testing energies for openff-2.0.0-rc.1...
================================================ warnings summary =================================================
openmmforcefields/tests/test_template_generators.py::TestGAFFTemplateGenerator::test_add_molecules
  /Users/mattthompson/software/openmmforcefields/openmmforcefields/utils.py:36: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
    from pkg_resources import resource_filename

openmmforcefields/tests/test_template_generators.py::TestGAFFTemplateGenerator::test_add_molecules
  /Users/mattthompson/micromamba/envs/interchange-examples-env/lib/python3.11/site-packages/pkg_resources/__init__.py:2832: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('google')`.
  Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
    declare_namespace(pkg)

openmmforcefields/tests/test_template_generators.py::TestSMIRNOFFTemplateGenerator::test_14_scaling_from_offxml
  /Users/mattthompson/micromamba/envs/interchange-examples-env/lib/python3.11/site-packages/lark/utils.py:163: DeprecationWarning: module 'sre_parse' is deprecated
    import sre_parse

openmmforcefields/tests/test_template_generators.py::TestSMIRNOFFTemplateGenerator::test_14_scaling_from_offxml
  /Users/mattthompson/micromamba/envs/interchange-examples-env/lib/python3.11/site-packages/lark/utils.py:164: DeprecationWarning: module 'sre_constants' is deprecated
    import sre_constants

openmmforcefields/tests/test_template_generators.py::TestSMIRNOFFTemplateGenerator::test_14_scaling_from_offxml
  /Users/mattthompson/micromamba/envs/interchange-examples-env/lib/python3.11/site-packages/mdtraj/formats/__init__.py:6: DeprecationWarning: 'xdrlib' is deprecated and slated for removal in Python 3.13
    from .xtc import XTCTrajectoryFile

openmmforcefields/tests/test_template_generators.py::TestSMIRNOFFTemplateGenerator::test_14_scaling_from_offxml
  /Users/mattthompson/micromamba/envs/interchange-examples-env/lib/python3.11/site-packages/mbuild/recipes/__init__.py:13: DeprecationWarning: SelectableGroups dict interface is deprecated. Use select.
    entry_points = metadata.entry_points()["mbuild.plugins"]

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
============================================= short test summary info =============================================
FAILED openmmforcefields/tests/test_template_generators.py::TestSMIRNOFFTemplateGenerator::test_energies - openmm.OpenMMException: HarmonicAngleForce: angle must be between 0 and pi
```

mattwthompson:
Improves molecule filtering in CI tests by

* Default to running 50 unique molecules in SMIRNOFF energy tests
* Default to filtering out molecules with > 40 atoms
* Randomly selecting the molecule(s) to run instead of always filtering out the same molecule

Might fix #330


mattwthompson:
Encouraged to see the expected failure: https://github.com/openmm/openmmforcefields/actions/runs/13993907851/job/39184133882?pr=370#step:6:499

epretti:
I found you may also need to add another exception to exclude from the list of forcefields.  `spce` should be excluded along with `tip` and `opc` in `TestSMIRNOFFTemplateGenerator.test_energies` (and maybe `TestGAFFTemplateGenerator.test_parameterize` too?).  I guess it was added at some point.  `spce` was excluded from only 1 of the 3 places where water models are checked and excluded.  Might be good to have a list at the top of the test file of the strings to check instead of duplicating.

Looks like there are some other actual failures in there in terms of energies not matching up:
https://github.com/openmm/openmmforcefields/actions/runs/13994740439/job/39186960028?pr=370#step:6:6353
I can take a look unless you're already digging into it.

epretti:
So just taking a look at the one molecule linked above that is failing, `[H][S][C](=[S])[N]([C]([H])([H])[C]([H])([H])[H])[C]([H])([H])[C]([H])([H])[H]`, the conformer generated by the OpenFF toolkit is far from the equilibrium geometry predicted by `smirnoff99Frosst-1.0.2`.  The large forces cause the molecule to disintegrate when dynamics are run, so one gets extremely large energy values.  They then still match very well between the reference and test systems, but the test is absolute energy difference instead of relative, so it fails.

This could be fixed by changing the energy difference check
https://github.com/openmm/openmmforcefields/blob/e6e6248af9dc1ce25303a7265969b116d0c220a4/openmmforcefields/tests/test_template_generators.py#L199-L201
to also include a relative as well as an absolute tolerance, and/or running a minimization between these lines
https://github.com/openmm/openmmforcefields/blob/e6e6248af9dc1ce25303a7265969b116d0c220a4/openmmforcefields/tests/test_template_generators.py#L1025-L1026
Perhaps that would fix some other failures as well.

The minimized geometry for this molecule looks extremely suspect to me.  The root issue for this particular molecule is that this force field is assigning a 4 angstrom (!) bond length.
Screenshot 2025-03-21 at 11 37 28

mattwthompson:
Heh, I was just in the process of tidying up a small number of details in that file before passing it off to you for a more serious analysis. I hoped to find an easy way to get `compare_energies` to report more information but found it wasn't something I could do very quickly. Anyway -


---

I think I speak for the entirety of the initiative in saying that smirnoff99Frosst was a essential stepping stone in the development of our force fields but was never meant to be used in perpetuity. Sage force fields mostly (but not completely) replace Parsley force fields which themselves mostly replaced smirnoff99Frosst. Skipping over some details, there's just no reason for somebody to use it as a "starting point" force field in the middle of 2025.

I don't think we should necessarily make it hard for people to use it in downstream packages like this, but it's also not worth the trouble of digging deep into its particular issues or putting a ton of CI time into testing it. I suggest skipping over it either in tests (my preference) or altogether.

epretti:
Happy to take a look at it whenever you're ready.  Thanks for the note on the context of the development of smirnoff99Frosst.

mattwthompson:
Seeing failures with (only) this molecule, so I'm looking to see if it's something that only happens with a particular force field

image

mattwthompson:
Okay, the failure is with 2.2.1RC1 but it's the same as the released force field:

```shell
$ diff ~/software/openff-forcefields/openforcefields/offxml/openff-2.2.1-rc1.offxml ~/software/openff-forcefields/openforcefields/offxml/openff-2.2.1.offxml
4c4
<     <Date>2024-07-22</Date>
---
>     <Date>2024-09-11</Date>
```

@epretti what should we do here?

epretti:
OK, I can reproduce, and I'm seeing this with a few force fields.  It's the same kind of problem as above with the other molecule I looked at.  In this case I think nothing's wrong with the force field, but the generated conformer is bad.  A close contact between two atoms creates large forces, the dynamics become unstable, and the resulting configuration can have extremely large energies.  Though they match to high precision in the test, the absolute value of the difference can be large.
image1
I think the best way to handle this (that appears to fix the problem for me) is to just minimize before running dynamics; changing
https://github.com/openmm/openmmforcefields/blob/e6e6248af9dc1ce25303a7265969b116d0c220a4/openmmforcefields/tests/test_template_generators.py#L1025-L1026
to
```python
context.setPositions(molecule.conformers[0].to_openmm()) 
openmm.LocalEnergyMinimizer.minimize(context)
integrator.step(nsteps) 
```

openmmforcefields/tests/conftest.py
```diff
@@ -25,6 +25,7 @@ def pytest_configure(config):
 
 
 def pytest_collection_modifyitems(config, items):
+    # why is this named skip_slow ??
```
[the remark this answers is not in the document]
mattwthompson: probably - I wanted to make note of it without trying to change it, since there's a small chance that would set me off on a wild goose chase that never ends

openmmforcefields/tests/test_template_generators.py
```diff
@@ -729,11 +774,9 @@ def test_jacs_complexes(self):
                 ligand_structures = [ligand_structures]
             assert len(ligand_structures) == len(molecules)
 
-            # Filter molecules
-            MAX_MOLECULES = 6 if not CI else 3
-            molecules = molecules[:MAX_MOLECULES]
-            ligand_structures = ligand_structures[:MAX_MOLECULES]
-            print(f"{len(molecules)} molecules remain after filtering")
+            ligand_structures = self.filter_molecules(molecules, max_molecules=3 if CI else 6)
```
epretti, line 777: This will cause a problem because the `ligand_structures` are ParmEd Structure objects, and not Molecule objects.  Adding them to Structure objects below, as `protein_structure + ligand_structure`, puts ParmEd into an infinite recursion.  This is a ParmEd bug as that should really raise TypeError, but in any case, the addition should be Structure + Structure.

mattwthompson: Whoops, didn't even know that was the type of that object. Only was reading portions of the code

mattwthompson: This should have caused some test(s) to fail

I think we can leverage an accident here in which these classes have similar and similarly-named attributes? ff351ba8329488959694bc57bb4825dd39a68dae


```python
In [1]: import parmed

In [2]: from openff.toolkit import Molecule

In [3]: molecule = Molecule.from_smiles(1000 * "C")

In [4]: structure = parmed.Structure()

In [5]: isinstance(molecule.atoms, list)
Out[5]: True

In [6]: isinstance(structure.atoms, list)
Out[6]: True

In [7]: len(molecule.atoms)
Out[7]: 3002

In [8]: len(structure.atoms)
Out[8]: 0

mattwthompson:
@epretti I think this is mostly ready for you to take over

epretti:
This should be working now.  I updated the troublesome GAFF partial charge test that we discussed recently to just create example user-defined charges that sum to the specified amount.  I'm creating a separate issue for what to do about this problem in general, and if it eventually leads to a PR, I can add separate tests there.

epretti (approved):

PR #375, opened 2025-04-08
set weights_only=False to ensure compatibility with torch>2.6

NiklasTR:
This is a minimal change that ensure torch.load of espaloma model checkpoints behaves as expected with torch 2.6+


mattwthompson (approved):
LGTM, a little surprised we haven't had this come up in the months since the 2.6 release

PR #386, opened 2025-06-24
Fixes and improvements for CHARMM force fields

epretti:
* Prevent proline-specific patch from being marked as applicable to non-proline residues (added test case covering this)
* Similar fixes for glycine and proline in Drude force field (added Drude version of test case also)
* Force field should now be usable with, *e.g.*, solvent FFXML files not specifically created with the atom type format used by `convert_charmm.py`: previously, improper/anisotropy assignment scripts would raise errors due to inability to parse the atom names
* Impropers/anisotropies for residues could be missed if the combination of patches that applied to a residue replaced *all*, not just some, of the residue's atoms: fixed this (new Drude test case covers this)
* Nicer formatting of lookup tables in improper/anisotropy scripts
* Update README to explain that custom patched version of ParmEd is needed for the conversion


epretti:
Yes, I should have linked back to openmm/openmm#4976.  Unfortunately the new CHARMM we released fails to parameterize systems with the existing water models that we didn't update (it raises an error instead), and the test in OpenMM that was loading the new force field doesn't use a system that has water in it, so I didn't catch it.

My plan after getting this merged into openmmforcefields was to open a PR for OpenMM that updates the force fields and adds a test that verifies that they work with the water models *in OpenMM*.  We should probably do a patch because the CHARMM36 2024 released now only works with the water models in openmmforcefields.

epretti:
@peastman If this looks OK, can you approve?  Although I have write access to this repository, the main branch is protected by a policy that doesn't let me merge unless another user explicitly approves the review.  I want to make sure this can be merged in to openmmforcefields and the main branch passes CI before openmm/openmm#4989 is merged.

peastman (approved):

PR #390, opened 2025-07-16
Add Lipid21

epretti:
Relevant changes in my patched branch of ParmEd are [`5592a24`](https://github.com/epretti/ParmEd/commit/5592a24d90590a20d2c999cd8326a44d4740667a) and [`98d5d5f`](https://github.com/epretti/ParmEd/commit/98d5d5ff4c95b3ba94f185f478a255d778893ec6).  This also affects GLYCAM because the script that ParmEd now uses to postprocess the exceptions is more general.


epretti:
Working on #393 has been awful in general (current state can be found at https://github.com/epretti/openmmforcefields/tree/validate-amber: this is still a work in progress!), but the lipid force fields have thankfully not contributed to the many problems that have come up.  So that we can get this merged here and then get Lipid21 into the OpenMM 8.4 beta, I've run some additional validation of this new force field.

In addition to the basic validation done by the existing conversion script, I've put together a lipid test suite in the `validate-amber` branch linked above to check OpenMM energies vs. sander.  All lipid head and tail groups supported by Lipid17 and Lipid21 should be covered, as well as all lipid residues in the "merged"/CHARMM-style files.  All tests pass for Lipid21 (and Lipid17 as well).  I've also checked that Lipid21 works with all of the Modeller-supported membranes, and that a basic membrane protein simulation with ff19SB + Lipid21 + OPC3 water can be parameterized and integrated stably for 100 ps.  This gives me more confidence that Lipid21 is OK to go.

The associated changes to ParmEd mentioned previously had caused the GLYCAM XML file to change as well.  None of the results should have changed, but I'm reverting that update in this PR for now until we can actually validate it properly by resolving #393.

PR #392, opened 2025-07-25
run espaloma tests

mikemhenry:


mikemhenry:
@j-wags Because we were not running the espaloma tests on CI, we missed that this if is also needed for the espaloma system generator. Instead of copying the fix over to the espaloma system generator, I opted to move the fix to `convert_system_to_ffxml` since  that is the method that ends up giving us the error with the `CMMotionRemover`. This way any future system generator templates that use ffxmls will get this fix as well.

epretti:
Thanks for getting these running!

It would be nice to test against OpenMM 8.3.1.  (Not sure if we would want to include it along with both 8.1.2 and 8.2.0, or drop 8.1.2 and keep testing only the most recent 2 minor releases.)  I had tried this in another PR I was working on and it was OK, but it might be worth seeing if this is possible here or if we'd run into any dependency conflicts with espaloma.

mikemhenry:
If we merge in https://github.com/openmm/openmmforcefields/pull/394 first then we can update this PR and check to see if everything still works okay.

mattwthompson:
I still see 3.10 checks, which I don't like to see

mikemhenry:
> I still see 3.10 checks, which I don't like to see

They are not running, the branch protection UI is being weird with what names it wants

I think there is a hole in the numpy/dgl/pytorch/python/openmm builds for python 3.13 with espaloma, once I confirm that I will set things up so we skip the espaloma tests for python 3.13

mikemhenry:
BTW, the root cause is there are no python 3.13 builds of DGL.

epretti:
Good to know.  I think it's fine then if we can make the espaloma test runs conditional on the Python version, as that would be better than not being able to run them at all.

epretti (approved):

PR #394, opened 2025-07-25
adjust ci matrix 

mikemhenry:
drop python 3.10, add 3.13 and drop openmm 8.1.2 and test 8.3.1


mattwthompson (approved):
Fine with force-merging given the quirk of changing requirements

PR #397, opened 2025-08-05
Remove `DummySystemGenerator`

### LINKED ISSUE #396: Remove `DummySystemGenerator`
See and several other comments in the same PR (search by `DummySystemGenerator` or similar keywords)

_Originally posted by @mikemhenry in https://github.com/openmm/openmmforcefields/issues/391#issuecomment-3137855583_

epretti:
Fixes #396; see #391 for context.  This class is broken and currently untested, so no changes other than removing it should be necessary here.


epretti:
There are some unusual errors that appear to be associated with espaloma and unrelated to this change: I will try to figure out what's happening.

PR #398, opened 2025-08-05
Fix espaloma model download race conditions

epretti:
This is an attempt to fix race conditions when multiple processes are trying to download espaloma models simultaneously (I believe this was the likely cause of test failures I saw in #397).  Specifically, if a process creates `ESPALOMA_MODEL_CACHE_PATH` in between another one checking that it doesn't exist and trying to create it, the second process would fail.  Also, if a process sees that the PyTorch model file exists, it would try to read it even if another process is still in the middle of downloading it, so it will run into an early EOF.

Here `makedirs(exist_ok=True)` should fix the first problem, and downloading to a temporary location before doing what should be an atomic rename should fix the second.  I suspect CI failures from these problems would be intermittent, but I can reliably reproduce them locally, and verify the fixes, by deleting `~/.espaloma` and running just the espaloma tests in parallel.

### DIFF
--- a/openmmforcefields/generators/template_generators.py
+++ b/openmmforcefields/generators/template_generators.py
@@ -1867,21 +1867,24 @@ class EspalomaTemplateGenerator(SmallMoleculeTemplateGenerator, OpenMMSystemMixi
                 return cached_filename
             else:
                 # Create the cache directory
-                if not os.path.exists(self.ESPALOMA_MODEL_CACHE_PATH):
-                    os.makedirs(self.ESPALOMA_MODEL_CACHE_PATH)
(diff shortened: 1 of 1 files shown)


mattwthompson:
We should consider using an off-the-shelf tool, @mikemhenry liked https://pypi.org/project/pooch/ in the past.

We (OpenFF) are facing issues with rate-limiting in CI when we hammer assets via GitHub Releases

epretti:
In an ideal world, this could be handled by espaloma itself.  It looks like there's [some capability](https://github.com/choderalab/espaloma/blob/main/espaloma/utils/model_fetch.py) for downloading models, but not handling automatic caching unless I missed it.  Is this fix to the current solution OK for now, though?

mikemhenry (approved):

PR #403, opened 2025-08-21
Using stdlib `importlib.resources` instead of `importlib_resources`

ijpulidos:
Just a minor nice to have. We could do the same without relying on the third party `importlib_resources`, and instead using the stdlib `importlib.resources` package.


ijpulidos:
To be precise, this is related to the issue https://github.com/conda-forge/openmmforcefields-feedstock/issues/25  in the feedstock.

PR #446, opened 2026-07-10
[DNM] Test Ambertools 25

mikemhenry:
I wanted to test ambertools 25 before I released it in some downstream envs, I only have uploaded the python 3.12 builds for linuxm, osx, and arm64 since I have limited space in my anaconda channel.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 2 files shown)


mikemhenry:
I also have no idea if the tests will pass/run/work on arm

peastman:
Looks like it's failing with a version conflict.  Espaloma 0.4.0 requires the existing OpenMMForceFields package, which requires an earlier version of AmberTools.

mikemhenry:
Yah espaloma is tricky with dgl being unmaintained, luckily testing ambertools 25 doesn't need espaloma to be in the env

PR #450, opened 2026-09-01
Update CHARMM36 to February 2026 release

epretti:
Updates CHARMM36 to the February 2026 release.  A few new problems with this release that required workarounds:
- The torsion `HB1-CT1-CT3-HA3` was added to `toppar/stream/prot/toppar_all36_prot_modify_res.str` to support a new N-methylalanine residue.  This overrides the torsion `X-CT1-CT3-X` when it is loaded, affecting regular alanine and therefore the energy of effectively every protein.  I am pretty sure this is a bug in the force field files that I will try to report/ask about.  With our current conversion system, the only way around this is to skip the entire file and all of its residues.
- Hydronium `H3O` was added, but instead of making a new atom type, they reused one from the *protein* force field, which has to be loaded with the water parameter files now for this to work.  This messes up the entire conversion system.  Unless someone really wants this, the easiest thing to do (that I did) is to skip this residue.
- The new release adds many new parameters, overflowing internal limits of old CHARMM versions, so a newer version is required.  The newest version (c51b1) is broken and not possible to compile without making modifications (see changeset).

This changes lots of force field files.  It's not clear if there are (intentional and/or unintentional) changes to existing residues, although there are definitely none in `charmm36_protein.xml`.  This should still probably be considered a new version of the force field, which raises the question of whether we want to replace the old one in OpenMMForceFields like we did last time or distribute both the 2024 and 2026 releases.


peastman:
It sounds like we should hold off on this until we get confirmation on whether those changes were intentional or bugs.

epretti:
I got a fast [response](https://forums-academiccharmm.org/viewtopic.php?p=40006) confirming that the torsion change is a bug.

I was also able to get CHARMM to compile without modifications by using CMake + Ninja with an option that their build script now provides.  I believe that the issue I was seeing was a CMake-related bug.

I'll rerun the conversion and test suite and see if any more issues show up.

epretti:
I've received confirmation on the forum that there will be a release with the fix, but not until February 2027.  Per our discussion, and since there hasn't been any particular demand for this release, we should probably hold off on this for now and revisit it later.

